PR #21, opened 2019-12-11
  Removed some cardinality restrictions related to Quantity

steveraysteveray:
Specifically, a Quantity can have multiple QuantityValues, each with a single number and unit.
e.g. a temperature reading could be 32 degrees F, AND 0 degrees C, should one want to explicitly say so.

Also, a Quantity and a QuantityValue can have multiple QuantityKinds, provided they are still commensurate. This would be useful when they are serving multiple communities, or within an interoperability exercise.

### DIFF
--- a/schema/SCHEMA_QUDT-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-v2.1.ttl
@@ -713,11 +713,6 @@ qudt:Quantity
       owl:minCardinality "0"^^xsd:nonNegativeInteger ;
       owl:onProperty qudt:belongsToSystemOfQuantities ;
     ] ;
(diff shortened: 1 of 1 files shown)


jhodgesatmb:
Steve,

You have gone over this before but I still do not understand Quantity. You say that QK is context free and that Q is context dependent, but the examples I have seen (Pressure vs AtmosphericPressure) don’t rise to that level in my opinion. To me AtmosphericPressure is a subclass of Pressure where you lock down an attribute value. If this makes it context sensitive then there would never be any subclassing using specialization. To me that is wrong.

So I need an explanation that makes sense to me. Is there one in the QUDT architecture narrative?

Jack

PR #28, opened 2019-12-17
  Added and populated new link qudt:omUnit for cross-reference to OM …

steveraysteveray:
…ontology

### DIFF
--- a/schema/SCHEMA_QUDT-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-v2.1.ttl
@@ -2436,6 +2436,10 @@ qudt:numericValue
   rdfs:label "numeric value" ;
   rdfs:range dtype:numericUnion ;
 .
+qudt:omUnit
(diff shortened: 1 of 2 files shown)


schema/SCHEMA_QUDT-v2.1.ttl
```diff
@@ -2436,6 +2436,10 @@ qudt:numericValue
   rdfs:label "numeric value" ;
   rdfs:range dtype:numericUnion ;
 .
+qudt:omUnit
+  a owl:ObjectProperty ;
+  rdfs:label "Links to the equivalent object in the OM ontology" ;
+.
 qudt:onlineReference
```
ralphtq: Should we not distinguish normative and informative references, as is done for standards? I think I already have properties for this.

steveraysteveray: I'm open to suggestions.

PR #33, opened 2020-01-17
Treatment of dimensionless quantity kinds

steveraysteveray:
I have added numerator and denominator dimension vectors for some **example** quantity kinds that are dimensionless ratios.
Specifically:
Angle
CompressibilityFactor
Gain
HeatCapacityRatio
MachNumber
MoleFraction
ReynoldsNumber
ThermalEfficiency

These examples will help with anyone tackling issue #34.


schema/SCHEMA_QUDT-v2.1.ttl
```diff
@@ -1442,12 +1442,12 @@ qudt:Unit
   rdfs:subClassOf [
       a owl:Restriction ;
       owl:allValuesFrom qudt:Unit ;
-      owl:onProperty qudt:hasDenominatorPart ;
+      owl:onProperty qudt:qkdvDenominatorPart ;
```
ralphtq: Do we need to append 'Part'?

schema/SCHEMA_QUDT-v2.1.ttl
```diff
@@ -1442,12 +1442,12 @@ qudt:Unit
...
     ] ;
   rdfs:subClassOf [
       a owl:Restriction ;
       owl:allValuesFrom qudt:Unit ;
-      owl:onProperty qudt:hasNumeratorPart ;
+      owl:onProperty qudt:qkdvNumeratorPart ;
```
ralphtq: Do we need to append 'Part'?

steveraysteveray: I thought about whether we should just call the property qkdvNumerator, without "Part", and decided to leave it in. But I could be convinced if you would like it removed. Also open to any other naming suggestion, before we start heavily using it.

Regarding pointing to a unit, I don't think so. Unit points to QuantityKind, which in turn points to QuantityKindDimensionVector.  (Jack was only temporarily linking some of the new IEC units to the dimension vectors in his branch).

schema/SCHEMA_QUDT-v2.1.ttl
steveraysteveray: I agree. It should not point to a Unit (and it doesn't). It points to a dimension vector.

steveraysteveray: Ah, I just noticed why you were talking about pointing to a Unit in the snippet above. Also noticed that you (or somebody) had already associated a different property with qudt:QuantityKind, namely qudt:numeratorDimensionVector (and one for denominator). I hadn't noticed that. Rather, I had just renamed the old qudt:hasDenominatorPart property, and neglected to even look at the schema definition for qudt:QuantityKind. That's my bad. 

So, since I need to go back and fix that, we might as well come to consensus on what we should call the property. Please voice your support for:
1. qudt:numeratorDimensionVector
2. qudt:numeratorQKDV
3. qudt:qkNumerator
4. Something else altogether.

Once we have consensus, I will fix the schema definition of qudt:QuantityKind, and update the small number of instances already populated.

steveraysteveray:
OK, I have now reset the qudt:hasDenominatorPart that was being used by a few qudt:Unit instances (and same for numerator).
Then, I renamed qudt:qkdvDenominatorPart to qudt:qkdvDenominator. This is used by the qudt:QuantityKind class.
Finally, I updated the 8 instances of qudt:QuantityKind with the renamed properties to the denominator and numerator dimension vectors.
I will also edit Issue #34 with these new names to avoid confusing readers.

PR #35, opened 2020-01-22
Added Geoscience file

nicholascar:
We would really like a dedicated class for all things Geoscience so I've created this file with the single `Geoscience Quantity Kind` class in it. Having it in its own file makes it easy to find and easier to manage than if it was merged in under science. We've versioned it "2.1" to match the other files, i.e. there isn't a v1.0, v2.0 etc.

I'm not sure if we will be adding more subclasses to this, but perhaps we might, looking at similar subclasses of `Quantity Kind`, such as `Atmospheric Quantity Kind` (4 subclasses) or `Biology Quantity Kind` (5 subclasses).

We will create individuals of types `Quantity`, `Quantity Kind` &`Quantity Kind Dimension Vector` in a separate submission which we will add not to the main file VOCAB_QUDT-QUANTITIES-v2.1.ttl but to a peer file within quantities/, VOCAB_QUDT-QUANTITIES-GEOSCIENCE-v2.1.ttl.

### DIFF
--- a/schema/SCHEMA_QUDT-GEOSCIENCE-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-GEOSCIENCE-v2.1.ttl
@@ -1,12 +1,15 @@
 # baseURI: http://qudt.org/2.1/schema/qudt/geoscience
 # imports: http://qudt.org/2.0/schema/qudt/science
 
+@prefix dc: <http://purl.org/dc/elements/1.1/> .
 @prefix dct: <http://purl.org/dc/terms/> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix qudt: <http://qudt.org/schema/qudt/> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix sdo: <https://schema.org/> .
 @prefix vaem: <http://www.linkedmodel.org/schema/vaem#> .
+@prefix voag: <http://voag.linkedmodel.org/schema/voag#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
 <http://qudt.org/2.1/schema/qudt/geoscience>
@@ -50,7 +53,6 @@ qudt:GMD_QUDT-SCHEMA-ENGINEERING
   vaem:isMetadataFor "http://qudt.org/2.1/schema/geoscience" ;
   vaem:revision "2.1" ;
   vaem:title "QUDT Schema for Geoscience" ;
-  vaem:usesNonImportedResource dct:abstract ;
   vaem:usesNonImportedResource dct:creator ;
   vaem:usesNonImportedResource dct:contributor ;
   vaem:usesNonImportedResource dct:created ;
@@ -62,9 +64,6 @@ qudt:GMD_QUDT-SCHEMA-ENGINEERING
   vaem:usesNonImportedResource dct:title ;
   vaem:usesNonImportedResource voag:QUDT-Attribution ;
   vaem:usesNonImportedResource <http://voag.linkedmodel.org/voag/CC-SHAREALIKE_3PT0-US> ;
-  vaem:usesNonImportedResource skos:closeMatch ;
-  vaem:usesNonImportedResource skos:exactMatch ;
-  vaem:usesNonImportedResource prov:wasInfluencedBy ;
   vaem:withAttributionTo voag:QUDT-Attribution ;
   rdfs:isDefinedBy <http://qudt.org/2.1/schema/qudt/geoscience> ;
   rdfs:isDefinedBy qudt:geoscience ;
--- a/schema/SCHEMA_QUDT-GEOSCIENCE-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-GEOSCIENCE-v2.1.ttl
@@ -1,4 +1,4 @@
-# baseURI: http://qudt.org/2.0/schema/qudt/geoscience
+# baseURI: http://qudt.org/2.1/schema/qudt/geoscience
 # imports: http://qudt.org/2.0/schema/qudt/science
 
 @prefix dct: <http://purl.org/dc/terms/> .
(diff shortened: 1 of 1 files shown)


nicholascar:
Discussion from now closed PR #36 moved here:

> concerning discipline-specific graphs, because you added the Geoscience file. This merits more discussion, since the same results can be achieved by either:
> 
> 1. Defining views for various disciplines, rather than augmenting the schema and vocabulary graphs,
> or
> 2. Building your own extensions to the QUDT ontology graphs for your own needs, keeping the QUDT ontologies discipline-neutral.
> 
> I'm not convinced we should add the Geoscience file here.
> Thoughts?

I prefer the first of the two methods you suggest for two reasons:

1. I can use the Profiles Vocabulary (PROF) to declare the profile
2. Keeping the profile in the main QUDT arena makes it available for others

If I do this, are you suggesting that the other disciplines like Finance will need to be moved into profiles or do you intend to keep the current top-level disciplines like Science, Finance etc and request profiling of those for sub disciplines such as geo-science?

If I do contribute a Geoscience discipline, should I create a PROF profile file within schema that declares the profile and lists the QUDT elements determined to be within the profile?

nicholascar:
Hi @steveraysteveray, could you respond the toe Q above about whether I should go ahead and make that profile? I now have a list of Units, Quantities etc. of relevance to the Geological Community, about 3/4 of which are already in QUDT, so I'm keen to follow the profiling approach to bundle them up. To do this I will bundle up the list of those geo Units, Quantities etc. in a new file, declare it a PROF profile of QUDT and submit it in place of this PR. Where should I aim to put that profile file? Let me know and I'll target the new PR there.

steveraysteveray:
We have discussed this, and definitely want to go the route of profiles (or views) to support various disciplines, including moving Finance, Science, etc. into profiles. We haven't yet come to a consensus about the use of the PROF vocabulary, which is why I had not yet responded. Will get back to you in a day or two, OK?

steveraysteveray:
@nicholascar, I could use a little education about the following. You mentioned at the start of this thread that you want something that will contain/link to "all things Geoscience". Using that as the example discipline, I am imagining that you want to list all the relevant Quantities, QuantityKinds and Units for the discipline in your chosen System of Units (e.g. SI or CGS). Could you walk me through the constituent classes and properties that you believe are needed, that point to the relevant Quantities, QuantityKinds and Units for that discipline? I'm not asking for every instance, but just examples, so that I can understand what is needed, and what is lacking, in the QUDT schema. We have a qudt:Discipline class that is bare bones right now, although we have stubbed out many (probably too many) instances of it, located in vocab/disciplines/VOCAB_QUDT-DISCIPLINES-v2.1.ttl. I'm wondering if we need much more than that, plus a few properties like: relevantQuantities, relevantQuantityKinds and relevant Units.
Would PROF be overkill?

nicholascar:
> We haven't yet come to a consensus about the use of the PROF vocabulary, which is why I had not yet responded. Will get back to you in a day or two, OK?

No problems. Please let me know if you want any help with PROF. To assist with this, I've created Issue #43.

> You mentioned at the start of this thread that you want something that will contain/link to "all things Geoscience". Using that as the example discipline, I am imagining that you want to list all the relevant Quantities, QuantityKinds and Units for the discipline in your chosen System of Units (e.g. SI or CGS). Could you walk me through the constituent classes and properties that you believe are needed, that point to the relevant Quantities, QuantityKinds and Units for that discipline?

As per the example in Issue #43, I would declare a profile (of QUDT All) for Geoscience that would include all the generic QUDT Quantities, QuantityKinds and Units thought relevant to geoscience and define additional derived units of geoscience-only interest.

I would then add the geoscience-specific derived units and perhaps include some other resources like examples.

I've created Issue #44 to pick up on this work


> Would PROF be overkill?

PROF's designed to allow specialised sub concerns of the overall domain do their own thing. For QUDT, I think it would usefully allow specialist communities to fiddle with their discipline's view of QUDT, creating derived units, guidance resources, extra validators etc., without polluting the base or generic QUDT. There are a lot of derived units in Geoscience, like *hundred revs per min* and *million barrels of oil equivalent per megalitre* that just don't need to be see by everyone!

Since the two questions are being addressed in Issues #43 & #44, I suggest closing this PR.

jhodgesatmb:
My concern is that units (etc.) that are relevant to many fields remain accessible to many fields and are not ‘only’ associated with one field. I don’t mind them being ‘marked’ in some way as associated with a field. I thought the issue here was one of trying to load appropriate units (etc) for runtime use, but my current understanding of what is wanted seems much more compartmentalized and could lead to duplication. Please tell me this is not the case and I am fine with it.

Jack

steveraysteveray:
I agree with @jhodgesatmb, in that it is hard to anticipate what units might be more generally useable/used by other disciplines, now or in the future, so I would rather not compartmentalize the declaration of the units themselves.

@nicholascar, why not keep **all** units in the same units graph where they can be discovered by others, and have the profile point to the ones of interest to that discipline? (Same for quantitykinds and quantities). That seems like a more sound approach.

And with that, I agree that we can continue this discussion in Issue #43.

jhodgesatmb:
I am very much in favor of @steverays suggestion, if it works with the scope of profiles.

Jack

dr-shorthair:
It might be worth thinking of a mechanism to **tag** or **annotate** units (and quantity kinds) with the  application domain, and allow the cardinality to be [0..*]

nicholascar:
@jhodgesatmb: "My concern is that units (etc.) that are relevant to many fields..."

Yes, I should be clearer: *all* QUDT units, as they currently are, would stay in one file, the profile declaration would just indicate those of interest to the profile's domain. This could be done by tagging them.

So yes, a good tagging mechanism would be sensible, rather that elements (Units, Quantities etc.) belonging to domain hierarchies.

The profiles would still be a good idea if a community wanted to do something more, like indicate non-standard equivalences. We would basically add geoscience quirks to a Geoscience profile that, if loaded, would import QUDT. So all the units would be available to all but not everyone would have to load the geoscience profile. Multi profiles could be added if that was useful for people.

jhodgesatmb:
Thank you for the clarification. It has always been a problem building semantic apps that load all of the units and quantities (and SOUs and SOQs) when you know your application domain doesn’t need them. The QUDT architecture has been designed to be true to Quantities, Units, Dimensions, and data Types, often at the expense of applications. It can be more than a repository and the use of Profiles, or something like them as Steve has proposed, might be a step in that direction.

PR #36, opened 2020-01-22
wholescale replacement of dct:author with dct:creator

nicholascar:
`dct:author` is not a property within Dublin Core Terms (see https://dublincore.org/specifications/dublin-core/dcmi-terms/). 

The closest term is `dct:creator`

### DIFF
--- a/schema/SCHEMA_QUDT-CITATION-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-CITATION-v2.1.ttl
@@ -85,7 +85,7 @@ dc:title
   vaem:title "QUDT Schema for Citations - Version 2.1" ;
   vaem:turtleFileURL "http://qudt.org/2.0/schema/SCHEMA_QUDT-CITATION-v2.1.ttl"^^xsd:anyURI ;
   vaem:usesNonImportedResource dcterms:abstract ;
-  vaem:usesNonImportedResource dcterms:author ;
+  vaem:usesNonImportedResource dcterms:creator ;
   vaem:usesNonImportedResource dcterms:created ;
   vaem:usesNonImportedResource dcterms:modified ;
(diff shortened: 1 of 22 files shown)


steveraysteveray:
Looks like you are correct. Too bad, because we have used dct:creator (I believe) to identify the original author, and dct:author as the principal contributor. (At least that was my interpretation). However, I can go along with the idea of collapsing the two terms into dct:creator.

However, this PR also relates to PR#35, concerning discipline-specific graphs, because you added the Geoscience file. This merits more discussion, since the same results can be achieved by either:
1. Defining views for various disciplines, rather than augmenting the schema and vocabulary graphs, 
or
2. Building your own extensions to the QUDT ontology graphs for your own needs, keeping the QUDT ontologies discipline-neutral.

I'm not convinced we should add the Geoscience file here.
Thoughts?

nicholascar:
> we have used `dct:creator` (I believe) to identify the original author, and `dct:author` as the principal contributor

So I think you should use `dct:creator` and `dct:contributor`, not that there's any sence of *principle* contributor just a/many contributors.

Yes, stupidly I added this PR on top of the geoscience work rather than making it independently so I guess it's now blocked by that! 

> ...the same results can be achieved by...

I prefer the first of the two methods you suggest for two reasons:

1. I can use the [Profiles Vocabulary](https://www.w3.org/TR/dx-prof/) (PROF) to declare the profile
2. Keeping the profile in the main QUDT arena makes it available for others

If I do this, are you suggesting that the other disciplines like Finance will need to be moved into profiles or do you intend to keep the current top-level disciplines like Science, Finance etc and request profiling of those for sub disciplines such as geo-science?

If I do contribute a Geoscience discipline, should I create a PROF profile file within schema that declares the profile and lists the QUDT elements determined to be within the profile?

PR #38, opened 2020-01-29
Fixauthor PR reissued

nicholascar:
A similar PR as before to fix dct:author but with no dependency on the Geo work. And dcterms/dct standardisation


dr-shorthair:
I've shifted (recently) to using `dcterms:` for the prefix, consistently, to avoid any potential confusion with `dctype:`.

PR #48, opened 2020-02-05
SOQK and SOU addition

steveraysteveray:


steveraysteveray:
Whoops, almost merged with master!

Issue #39, opened 2020-01-29
Make qudt:Concept sub-class-of skos:Concept


dr-shorthair:
QUDT individuals can be reasonably understood to behave as a special kind of 'concept' in the sense designed in SKOS. Conceived this way, sets of QUDT individuals could be managed and published through standard vocabulary services. 

All that is needed is to add the axiom
```
qudt:Concept rdfs:subClassOf skos:Concept .
```

nicholascar:
Could QUDT consider using `skos:Concept` directly instead of subclassing?

steveraysteveray:
As I understand it, there are many entailments for a skos:Concept, which may or may not be appropriate for QUDT individuals. I need to study up more on skos.

dr-shorthair:
Not really. And none that would get in the way, that I have encountered.

nicholascar:
Agreed: no problematic entailment for skos:Concept that I can think of, indeed there are hardly any entailments at all. There are no mandatory properties, whole/part relations that place Concepts in relation to other containers, etc.

Concepts really just indicate the thing is generically an abstract *concept* and that certain annotation properties - the SKOS labels - can be used and other object properties could be used to relate it to other concepts in a hierarchy.

nicholascar:
See PR #57

Issue #43, opened 2020-02-04
Use the PROF vocabulary to to supply profiles


nicholascar:
Instead of using keeping the various disciplines, Finance, Science, etc. just in separate file, they could be declared as profiles of QUDT using the W3C's [Profiles Vocabulary](https://www.w3.org/TR/dx-prof/). 

Anything that conforms to a Profile, *MUST* conform to the QUDT base specification while a profile may also implement additional constraints on the base specification.

This would look something like this:

*(the profile declaration file for Finance)*
```
@prefix prof: <http://www.w3.org/ns/dx/prof/> .  # the Profiles Vocabulary
@prefix role: <http://www.w3.org/ns/dx/prof/role/> .  # the vocabulary of profile part roles

<http://qudt.org/2.0/schema/qudt/finance>
  a owl:Ontology ;  # this is still an ontology
  a prof:Profile ;     # additionally it's a profile, of the main QUDT
  prof:isProfileOf <http://qudt.org/2.1/schema/qudt> ;  # QUDT identified by namespace URI
  rdfs:label "QUDT Schema - Finance" ;
  owl:imports <http://qudt.org/2.1/schema/qudt> ;  # owl:imports owl:subPropertyOf prof:isProfileOf 
  prof:hasResource [
    a prof:ResourceDescriptor ;  # this is one of the elements of the profile
    prof:hasRole role:vocabulary ;  # the role of this ResourceDescriptor
    prof:hasArtifact <SOME-FILE-IN-THE-QUDT-REPO> ;  # the actual vocab file
  ] ;
  prof:hasResource [
    a prof:ResourceDescriptor ;  
    # other ResourceDescriptors, if relevant
  ] ;
...
.
```
In this example, QUDT All contains all of QUDT and this profile for Finance imports it and then extends on it with any new elements that it declares within the vocabulary resource indicated. If the Finance sector has additional validation rules, guidance documents or examples, additional `ResourceDescriptors` can be described.

The purpose of doing this the PROF way is that now all of generic QUDT is in one place and Finance just declares the extra, finance-only, bits.

nicholascar:
I volunteer to profile-ise QUDT disciplines, if the consortium votes for this.

steveraysteveray:
I definitely prefer "outside pointing in" (i.e. profiles/views) to "inside pointing out" (i.e. tags), because as the number of disciplines grows, a given unit would have gobs of tags attached to it, with all but one of them of no relevance to a given application.

[...]

PR #57, opened 2020-02-25
qudt:Concept as per Issue 39

nicholascar:
As per https://github.com/qudt/qudt-public-repo/issues/39

### DIFF
--- a/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
@@ -1415,7 +1415,7 @@ qudt:Concept
   rdfs:label "QUDT Concept" ;
   rdfs:subClassOf owl:Thing ;
   rdfs:subClassOf skos:Concept ;
-  skos:definition "An QUDT idea or notion; a QUDT unit of thought."@en .
+  skos:definition "An QUDT idea or notion; a QUDT unit of thought."@en ;
   skos:scopeNote "QUDT individuals can be reasonably understood to behave as a special kind of 'concept' in the sense designed in SKOS. Conceived this way, sets of QUDT individuals could be managed and published through standard vocabulary services and the subclassing of qudt:Concept as a skos:Conceptwill allow for SKOS-specific tooling to 'see' them."@en ;
 .
 qudt:Conference
--- a/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
@@ -1414,6 +1414,9 @@ qudt:Concept
   a sh:NodeShape ;
   rdfs:label "QUDT Concept" ;
   rdfs:subClassOf owl:Thing ;
+  rdfs:subClassOf skos:Concept ;
+  skos:definition "An QUDT idea or notion; a QUDT unit of thought."@en .
+  skos:scopeNote "QUDT individuals can be reasonably understood to behave as a special kind of 'concept' in the sense designed in SKOS. Conceived this way, sets of QUDT individuals could be managed and published through standard vocabulary services and the subclassing of qudt:Concept as a skos:Conceptwill allow for SKOS-specific tooling to 'see' them."@en ;
 .
 qudt:Conference
   a qudt:PubEnumeratedType ;


steveraysteveray:
We're still discussing.
Meanwhile, of course, you can just have an "overlay" file in your workspace with this declaration.

steveraysteveray:
We discussed some more. The issue is the implication of applying skos behavior on concepts that aren't really terminological. The main exception is QuantityKind, which IS terminological, even though it also has object properties. 
In other words, what would it mean to say one Unit instance is skos:broader than another?
On the other hand, this makes sense for QuantityKind, where, say, Energy is broader than MechanicalEnergy.
How would you feel about only declaring qudt:QuantityKind as a subClass of skos:Concept, rather than all qudt:Concept classes being so declared?

dr-shorthair:
You clearly have a view on the meaning of 'terminological' which is stronger than mine! 
AFAIK `skos:Concept` is effectively a weak form of `rdfs:Class`. 
What is `qudt:Concept`?

steveraysteveray:
My understanding is that a skos:Concept is a "term" that can be applied to an actual "concept". 
qudt:Concept is itself a "concept" in the sense of conceptual modeling.
https://terms.tdwg.org/wiki/Term_versus_Concept

dr-shorthair:
`skos:Concept` is precisely the same thing - it is an abstract notion, with one or more preferred and alternative labels per natural language. The URI identifies the concept, not the label. i.e. the concept comes first, not the term.  https://www.w3.org/TR/skos-primer/#secconcept

steveraysteveray:
Yes, I see what you are referring to.  However, I'm still asking the question about Unit versus QuantityKind above (https://github.com/qudt/qudt-public-repo/pull/57#issuecomment-607972036). I don't know how one Unit can be broader or narrower than another, while I CAN understand that for QuantityKinds.

dr-shorthair:
It is not mandatory to incorporate `skos:broader` and 'skos:narrower` assertions in your vocabulary. These are convenience properties, to be used when the subject matter matches, but not otherwise.

dr-shorthair:
I would see `qudt:generalization rdfs:subPropertyOf skos:broader.` but that does not imply that skos:broader shoudl be used on non-QuantityKind indidivduals.

steveraysteveray:
In the end, we discussed this again and have concluded that we will not include the "qudt:Concept rdfs:subClassOf skos:Concept" assertion in the QUDT schema, primarily because we cannot anticipate its impact on our other users. It's a global change.

dr-shorthair:
Ok - we are looking at the same issue (SKOS is a widely adopted schema) and coming to opposite conclusions 
- I see the benefits in aligning and leveraging SKOS services and infrastructure, 
- you see alignment as introducing risks

I understand that premature alignment can be seen as an anti-pattern, and that adopting SKOS could be seen as introducing some poorly defined semantics into a more strongly axiomatized framework. But could I suggest adding an alignment graph, as a separate file in the QUDT distribution, so people who would like to leverage the alignment have some guidance here? It would really only need two axioms:
```
qudt:Concept rdfs:subClassOf skos:Concept .
qudt:generalization rdfs:subPropertyOf skos:broader. 
```

steveraysteveray:
Interesting suggestion. I'll bring it up at the next Board meeting. Perhaps we could include a discussion of this in the documentation, such as on the wiki pages.

steveraysteveray:
Nicholas & Simon,
I think we have a solution that will satisfy all of us. Please see https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#customizing-the-qudt-ontology and PR #104.

dr-shorthair:
None of the links work on the Wiki page :-(

nicholascar:
I think Simon’s referring to the fact that when we visit the web addresses on that wiki page, http://qudt.org/2.1/schema/extensions/imports & http://qudt.org/2.1/schema/extensions/skos, we see a 404: Page not found.

The "Advanced User Guide" webpage itself works!

By the way, I’m about to start serious work on the geoprofile of QUDT since the Department I work for has collated all the Units they use in the geoscience area.

steveraysteveray:
Ah, thanks for the clarification. It's true that those URIs don't resolve, but they aren't intended to. They are just the base URIs of the two graphs in question.

Great news on the geo-profile. Issue #43 is still open.

PR #78, opened 2020-03-30
Rh qudt shacl validation

ralphtq:


steveraysteveray:
Ralph,
While looking this over, I noticed that all (I didn't check for all) the restrictions now contain:
      rdfs:isDefinedBy <http://qudt.org/2.1/schema/qudt> ;

I'm thinking this might have been caused when you were adding that to normal class definitions. Seems a bit excessive for an anonymous class restriction. Could you remove those?

schema/SCHEMA_QUDT-v2.1.ttl
```diff
@@ -1047,24 +1047,12 @@ qudt:Quantity
       owl:allValuesFrom qudt:QuantityValue ;
       owl:onProperty qudt:quantityValue ;
     ] ;
-  rdfs:subClassOf [
```
steveraysteveray: We still want Quantity to point to QuantityKind.
We don't want QuantityValue to point to QuantityKind.

ralphtq: This is done

ralphtq:
@steveraysteveray  - are we done now on this PR?

schema/SCHEMA_QUDT-v2.1.ttl
steveraysteveray: I'm not seeing Quantity pointing to QuantityKind.

steveraysteveray:
I don't think so. Quantity needs to point to QuantityKind.

PR #96, opened 2020-04-15
Added LaTeX meta-characters to latexSymbol values

ralphtq:


jhodgesatmb:
I am going to stop doing IEC unit additions as these latex changes are causing conflicts with my additions.

PR #104, opened 2020-04-22
Extension capability to support external ontologies

steveraysteveray:


schema/SCHEMA_QUDT-v2.1.ttl
```diff
@@ -1,4 +1,5 @@
 # baseURI: http://qudt.org/2.1/schema/qudt
+# imports: http://qudt.org/2.1/schema/extensions/imports
```
ralphtq, line 2: propose http://qudt.org/2.1/schema/qudt/extensions

rationale: 

1. same as qudt schema base uri up to /extensions
2. extensions can be more than graph imports - if people choose to do that

schema/extensions/skosext.ttl
```diff
@@ -0,0 +1,23 @@
+# baseURI: http://qudt.org/2.1/schema/extensions/skosext
```
ralphtq, line 1: propose http://qudt.org/2.1/schema/qudt/extensions/skos

rationale:

same as qudt schema base uri up to /extensions/skos

jhodgesatmb:
Agreed. Also, extensions and imports could be construed as redundant by some.

Jack Hodges, Ph.D.
Arbor Studios

schema/extensions/skosext.ttl
steveraysteveray: The reason I didn't call it skos is because the intuitive prefix for that would then be "skos", but that collides with the prefix for the real skos. Of course, we can still define the prefix to be skosext anyway, but it kind of breaks with convention.
Thoughts?

schema/SCHEMA_QUDT-v2.1.ttl
steveraysteveray: Because extensions can be more than just imports is why I called this one "imports", allowing the existence of other things in the directory as well.

schema/extensions/skosext.ttl
steveraysteveray: Agreed to the URI being http://qudt.org/2.1/schema/qudt/extensions/skos
If we choose to define a prefix, it could be something like qudt.skos or qudt-skos.

schema/SCHEMA_QUDT-v2.1.ttl
steveraysteveray: Agreed that this imports file/URI will only contain import statements.
An example import would be the qudt skos extension file.

steveraysteveray:
OK, all changes we discussed have been made. Ready for another review.

dr-shorthair:
It would be simpler if schema/extensions/skos.ttl just imported QUDT directly. Then people could refer to this graph and would not have to do any editing. The latter assumes that folk will download or cache the repo, while the former strategy allows them to leave everything in place.

steveraysteveray:
There are two reasons we did not do as you suggest.
1. Our approach allows the user to include combinations of extensions, of which skos is just the first example. We anticipate there will be many others, in which case the user would add import statements for their particular needs.
2. As I'm sure you know, members of the semantic technology community tends to adopt one of two approaches: a) the linked data approach where all graph, class and property URIs resolve into URLs on the web; and b) the "industrial use" approach where all the ontologies used in an application are maintained under careful local version control in order to achieve predictable behavior. There are good reasons for each of them. We are striving to provide QUDT in a way that is useful for either approach, but some of the design decisions may at times be a little awkward 
 for one or the other.

steveraysteveray:
Simon, after some discussion we agreed that there is no harm in having skos.ttl import QUDT directly, for the reasons you suggested. This way, the linked model world can use it, while the industrial use community can adopt the procedure as documented in the [advanced user guide](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#customizing-the-qudt-ontology). So, PR #114 does this.

dr-shorthair:
Thanks for entertaining constructive criticism.

jhodgesatmb:
What did I say this time?

Jack Hodges, Ph.D.
Arbor Studios

dr-shorthair:
All good.  Thank _you_ for entertaining _my_ constructive criticism.

Issue #108, opened 2020-04-23
M3 description incorrect


dr-shorthair:
http://qudt.org/vocab/unit/M3

"The SI unit of volume, equal to **106 cm3**, 1000 liters, 35.3147 ft3, or **1.307 95 yd3**. "

The items in **bold** are incorrect.

steveraysteveray:
Thanks Simon. Hadn't examined the description. I agree with the cm3 erratum, but not the yd3. Here's the output from my handy conversion SPARQL query:

To convert	Cubic Meter	into	Cubic Mile	multiply by	2.39912758316496E-10
To convert	Cubic Meter	into	MegaL	multiply by	1.0E-6
To convert	Cubic Meter	into	Acre Foot	multiply by	8.107131937899124E-4
To convert	Cubic Meter	into	DecaM3	multiply by	0.001e0
To convert	Cubic Meter	into	Standard	multiply by	0.21404109589041098e0
To convert	Cubic Meter	into	Cord	multiply by	0.27624309392265195e0
To convert	Cubic Meter	into	Register Ton	multiply by	0.3531466672148859e0
To convert	Cubic Meter	into	TON_SHIPPING_US	multiply by	0.8829242450997704e0
To convert	Cubic Meter	into	KiloL	multiply by	1.0e0
To convert	Cubic Meter	into	Stere	multiply by	1.0e0
To convert	Cubic Meter	into	Cubic Yard	multiply by	1.3079506193143922e0
To convert	Cubic Meter	into	BBL_UK_PET	multiply by	6.284833690730875e0
To convert	Cubic Meter	into	BBL_US	multiply by	6.28981056977507e0
To convert	Cubic Meter	into	BBL_US_DRY	multiply by	8.6484093756776e0
To convert	Cubic Meter	into	HectoL	multiply by	10.0e0
To convert	Cubic Meter	into	bushel (UK)	multiply by	27.496156037385973e0
To convert	Cubic Meter	into	bushel (US)	multiply by	28.377593392788178e0
To convert	Cubic Meter	into	Cubic Foot	multiply by	35.31466672148859e0
To convert	Cubic Meter	into	DecaL	multiply by	100.0e0
To convert	Cubic Meter	into	PK_UK	multiply by	109.98461205292769e0
To convert	Cubic Meter	into	GAL_UK	multiply by	219.96924829908778e0
To convert	Cubic Meter	into	Imperial Gallon	multiply by	219.96924829908778e0
[...]

PR #109, opened 2020-04-23
 Fixed description of M3

steveraysteveray:


dr-shorthair:
Fixes #108

steveraysteveray:
Nice, Simon. A more elegant way of closing the issue!

dr-shorthair:
Hmm - still showing 1m^3 is 106cm^3 - http://qudt.org/vocab/unit/M3 
Something in the rendering pipeline?

steveraysteveray:
Ah yes. This PR was merged on 4/24/2020, but the html and ttl pages on the web were last regenerated on 4/18/2020. It takes several hours to regenerate, which I only do every few weeks.  Maybe I should bite the bullet and do another round.

PR #110, opened 2020-04-24
Added a few units, modified some errors in already-submitted units, a…

jhodgesatmb:
…nd added one quantity kind and one dimension vector


vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -8693,6 +8693,15 @@ For an irrotational electric field, the voltage is independent of the path betwe
   rdfs:label "Voltage" ;
   skos:broader quantitykind:EnergyPerElectricCharge ;
 .
+quantitykind:VoltageLoss
```
steveraysteveray, line 8696: I don't think this is right. The dimension of Voltage Loss is the same as Voltage, namely ML2T-3I-1.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -237,6 +237,18 @@ unit:A-SEC
   rdfs:label "A SEC" ;
   skos:prefLabel "ampere second" ;
 .
+unit:A2-HR
```
steveraysteveray, line 240: I don't understand this unit. The unit for measuring a Voltage Loss is the Volt.

vocab/dimensionvectors/VOCAB_QUDT-DIMENSION-VECTORS-v2.1.ttl
```diff
@@ -2546,6 +2546,23 @@ qkdv:A0E2L0I0M-1H0T4D0
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/dimensionvector> ;
   rdfs:label "A0E2L0I0M-1H0T4D0" ;
 .
+qkdv:A0E2L0I0M0H0T1D0
```
steveraysteveray, line 2549: No problem with this dimension vector, but it has nothing to do with Voltage or Voltage Loss as far as I can tell.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -14969,6 +15010,15 @@ unit:OHM-M
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Ohm Meter" ;
 .
+unit:OHM-M2-PER-M
```
steveraysteveray, line 15013: I suppose this is OK. We already have the OHM-M as a unit of Resistivity.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -15060,7 +15110,7 @@ unit:OZ_M-FT
   qudt:iec61360Code "0112/2///62720#UAB133" ;
   qudt:uneceCommonCode "4R" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
-  rdfs:label "OZ_M_US FT" ;
+  rdfs:label "OZ_M FT" ;
```
steveraysteveray, line 15113: Something wrong here. The unit is OHM_Ab, and the label is OZ_M.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -15245,7 +15295,7 @@ unit:OZ_VOL_UK-PER-DAY
   qudt:iec61360Code "0112/2///62720#UAA432" ;
   qudt:uneceCommonCode "J95" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
-  rdfs:label "OZ_UK PER DAY" ;
+  rdfs:label "OZ_VOL_UK PER DAY" ;
```
steveraysteveray, line 15298: The label doesn't match the qname.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -16197,6 +16257,16 @@ unit:PERCENT-OHM-PER-OHM
   rdfs:label "PERCENT OHM PER OHM" ;
   skos:prefLabel "percent ohm per ohm" ;
 .
+unit:PERCENT-SEC-PER-SEC
+  a qudt:Unit ;
+  qudt:conversionMultiplier 1.0e-00 ;
```
steveraysteveray, line 16262: Using the same logic as with PERCENT-MI-PER-MI, shouldn't this conversionMultiplier also be 1e-02? Basically just converting a percent into a decimal, right?

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -19495,6 +19565,15 @@ unit:V-SEC-PER-M
   rdfs:label "V SEC PER M" ;
   skos:prefLabel "volt second per metre" ;
 .
+unit:V2-HR
+  a qudt:Unit ;
+  qudt:conversionMultiplier 3.6e03 ;
+  qudt:conversionOffset "0.0"^^xsd:double ;
+  qudt:expression "v^2hr" ;
+  qudt:hasQuantityKind quantitykind:LorenzCoefficient ;
```
steveraysteveray, line 19573: How can this unit have the same quantity kind as unit:V2-PER-K2? Something is wrong.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: I knew that you would take issue with this but I did find a reference to voltage loss for the unit that said that voltage loss is measured in ampere-squared-hour (or ampere-squared-second). Now I am looking for the reference. In the meantime, I found a couple of other references to it that are both similar and different:

https://grabcad.com/library/ampere-squared-hour-meter-1
BACnet - Ampere square hours (246) ???

The ampere-squared-seconds is a unit measure based on the energy or heat allowed to pass through the fuse or circuit breaker during a short-circuit condition.  The unit is derived by squaring the effective RMS current and multiplying it by the total time it takes the fuse or circuit breaker to open and extinguish the electrical arcing.  The total clearing ampere-squared-seconds (I²tc) is the sum of the “opening” ampere-squared-seconds (I²to) plus “arcing” ampere-squared-seconds (I²ta).  (I²tc = I²to + I²ta).

https://www.optifuse.com/glossary.php

The DV associated with VoltageLoss is simply the units for A2-HR, so if VoltageLoss is not the quantity kind for A2-HR then we fix that and we are ok.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: Please see the references I gave. I am glad that you take issue with these because I have no basis to question them; I just look them up and if I find references then I no longer question the unit but look for an understanding of its quantity kind.

One thing I found out is that some people confuse Voltage Loss with Voltage Drop and that they aren't the same.

vocab/dimensionvectors/VOCAB_QUDT-DIMENSION-VECTORS-v2.1.ttl
jhodgesatmb: Then lets fix the A2-HR reference and quantity kind.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: I know ...

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: I must have missed this. I took the DV for Volt: A0E-1L2I0M1H0T-3D0. If I square this then I would double the exponents wouldn't I? That would make: A0E-2L4I0M2H0T-6D0. If I then multiply by T1 I should get: A0E-2L4I0M2H0T-5D0. There is no such dimension vector in QUDT (for V2 'or' V2-HR).

Here is a reference:

https://patents.google.com/patent/US4066959A/en

In the metering of electrical energy it is often advantageous to know the number of volt-square-hours (V2 hr) as, for example, in the measurement of apparent electrical power and energy. Apparent electrical power is the product of current and voltage, oftentimes expressed in kilovolt-amperes (kVa), and is independent of the phase angle between the current and voltage. Apparent electrical energy is expressed in kilovolt-ampere-hours (kVah) which is also independent of the phase angle between the current and the voltage. As is known in the art, both apparent electrical energy and apparent electrical power are proportional to V2 hr and V2, respectively. In many instances, therefore, a simplified way of determining apparent electrical power or energy in a system is to determine the number of V2 hrs.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: All of the OZ_VOL_UK qnames match their labels. What unit are you referring to? I do not see it here.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: Steve,

I am only seeing OHM_Ab for OHM_Ab, and I only see OZ_M FT for OZ_M-FT

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: I suggest we put these questionable quantity kinds into a holding pen until we can bring on a subject matter expert to help us resolve them. My intuition tells me that the dimensionality should be the same as voltage, but I don't know what I don't know.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
steveraysteveray: Recommend the "holding pen" for this.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -4252,6 +4264,35 @@ unit:DeciB_M
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Decibel Referred to 1mw" ;
 .
+unit:DeciB_MilliV
+  a qudt:DimensionlessUnit ;
+  a qudt:LogarithmicUnit ;
+  a qudt:Unit ;
+  dcterms:description "<p class=\"lm-para\">It can be used to express a change in value (e.g., +1 dB or −1 dB) or an absolute value. In the latter case, it expresses the ratio of a value to a fixed reference value; when used in this way, a suffix that indicates the reference value is often appended to the decibel symbol. For example, if the reference value is 1 volt, then the suffix is \"V\" (e.g., \"20 dBV\"), and if the reference value is one milliwatt, then the suffix is \"m\" (e.g., \"20 dBm\").</p>"^^qudt:LatexString ;
```
steveraysteveray, line 4271: This description still seems illogical. 
if the reference value is 1 volt, then the suffix is \"V\" (e.g., \"20 dBV\") - OK, that sounds good.
if the reference value is one milliwatt, then the suffix is \"m\" (e.g., \"20 dBm\") - Wait! Why "m" and not "mW"?
How would we handle millivolt if "m" is used for milliwatt?

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -4252,6 +4264,35 @@ unit:DeciB_M
...
+  qudt:informativeReference "https://en.wikipedia.org/wiki/Decibel"^^xsd:anyURI ;
+  qudt:symbol "dBm" ;
+  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
+  rdfs:label "Decibel MilliVolt reference" ;
+.
+unit:DeciB_Volt
+  a qudt:DimensionlessUnit ;
+  a qudt:LogarithmicUnit ;
+  a qudt:Unit ;
+  dcterms:description "<p class=\"lm-para\">It can be used to express a change in value (e.g., +1 dB or −1 dB) or an absolute value. In the latter case, it expresses the ratio of a value to a fixed reference value; when used in this way, a suffix that indicates the reference value is often appended to the decibel symbol. For example, if the reference value is 1 volt, then the suffix is \"V\" (e.g., \"20 dBV\"), and if the reference value is one milliwatt, then the suffix is \"m\" (e.g., \"20 dBm\").</p>"^^qudt:LatexString ;
```
steveraysteveray, line 4286: See my comment on DeciB_MilliV

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -16157,6 +16229,16 @@ unit:PERCENT-M-PER-M
   rdfs:label "PERCENT M PER M" ;
   skos:prefLabel "percent metre per metre" ;
 .
+unit:PERCENT-MI-PER-MI
+  a qudt:Unit ;
+  qudt:conversionMultiplier "0.000006215040398"^^xsd:double ;
```
steveraysteveray, line 16234: Why not 0.01?

steveraysteveray: I'm not understanding why all these dimensionless ratios don't just have 0.01 as the conversion multiplier.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -19491,6 +19583,15 @@ unit:V-SEC-PER-M
   rdfs:label "V SEC PER M" ;
   skos:prefLabel "volt second per metre" ;
 .
+unit:V2-HR
```
steveraysteveray, line 19586: As with Ampere-squared-hour, I recommend we put in a holding pen and seek out an expert.

jhodgesatmb:
That is a problem, as I did not create that symbol. Also Steve, I think you misread because it is MilliVolt not MilliWatt. The problem is the same though because dBm is no more millivolt than it is milliwatt. Maybe it was mv and I made a mistake. I’ll check.

jhodgesatmb:
Regarding the percent-foo-per-foo it is because the numerator is implied, so we only use the foo conversion multiplier in the denominator. We did talk about this. Originally the unit was just percentage-per-foo, but that didn’t make sense as a percentage. We got a sample from Maja’s husband that cleared it up. So the percentage gives us the .01 but we multiply that by the percent-per-foo conversion value. I am fine putting all these units into a holding cell though. We want to do what is right.

Jack

steveraysteveray:
> Regarding the percent-foo-per-foo it is because the numerator is implied, so we only use the foo conversion multiplier in the denominator. We did talk about this. Originally the unit was just percentage-per-foo, but that didn’t make sense as a percentage. We got a sample from Maja’s husband that cleared it up. So the percentage gives us the .01 but we multiply that by the percent-per-foo conversion value. I am fine putting all these units into a holding cell though. We want to do what is right. Jack


I'm not yet convinced that the conversion multiplier shouldn't be 0.01.
 - Steve

steveraysteveray:
> That is a problem, as I did not create that symbol. Also Steve, I think you misread because it is MilliVolt not MilliWatt. The problem is the same though because dBm is no more millivolt than it is milliwatt. Maybe it was mv and I made a mistake. I’ll check.

The description for DeciB_MilliV talks about MilliWatt as an example, as I recall.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -16135,7 +16207,7 @@ unit:PERCENT-HectoBAR-PER-HectoBAR
 .
 unit:PERCENT-IN-PER-IN
   a qudt:Unit ;
-  qudt:conversionMultiplier 3.93701e-01 ;
+  qudt:conversionMultiplier "0.393700787401575"^^xsd:double ;
```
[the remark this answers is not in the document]
steveraysteveray: Jack, I think this should be treated similarly to the new treatment of PERCENT_RH, namely with a quantitykind of DimensionlessRatio, and with a contextualquantitykind of LengthPercentage. Let me know if you'd like me to make these changes.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -16171,7 +16288,7 @@ unit:PERCENT-MO-PER-MO
 .
 unit:PERCENT-MilliM-PER-MilliM
   a qudt:Unit ;
-  qudt:conversionMultiplier 1.0e-02 ;
+  qudt:conversionMultiplier "10.0"^^xsd:double ;
```
steveraysteveray, line 16291: Same comments as for PERCENT-IN-PER-IN; conversionMultiplier, and the two quantitykind pointers.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -6212,6 +6212,16 @@ unit:GM-PER-DeciM3
   rdfs:label "GM PER DeciM3" ;
   skos:prefLabel "gram per decimetre cubed" ;
 .
+unit:GM-PER-GM
```
steveraysteveray, line 6215: I think this should be treated according to our discussion, namely with a quantitykind of DimensionlessRatio, and with a contextualquantitykind of MassRatio. Let me know if you'd like me to make these changes.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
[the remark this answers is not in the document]
jhodgesatmb: That is ok, but the conversion multiplier should remain as it is.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: Because the numerator is implied in percent, but it was really percent-per-MI, so it would 0.01 * 1/(CM for MI). We renamed them from PERCENT-PER-<FOO> because that would seem like it isn't a percentage at all.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
jhodgesatmb: The example you are showing, Steve, is GM-PER-DeciM3, which is not a MassRatio but a MassDensity, isn't it?

steveraysteveray: My comment is attached to unit:GM-PER-GM.

steveraysteveray: image

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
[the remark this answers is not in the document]
steveraysteveray: We'll need to discuss further. I disagree.

PR #119, opened 2020-05-13
_Metric -> _M

nicholascar:
Inconsistent use of _M or _Metric with TON (tonne) so I've aligned all with _M as used on the base unit.


nicholascar:
> we should resolve the inconsistency

Great. Feel free to close this PR if you address this elsewhere.

PR #122, opened 2020-05-15
Finished up units lacking a quantity kind

steveraysteveray:
...except for these 3, which I am a bit suspicious of:

Units that are missing Quantity Kinds
unit:PER-SEC-M2
unit:PER-SEC-M2-SR
unit:PER-SEC-SR

(These have now been fixed in subsequent commits of this PR)


dr-shorthair:
`unit:PER-SEC-M2` and `unit:PER-SEC-SR` look like fairly standard particle-flux units. 
Not sure about the other one.

steveraysteveray:
Interesting. I'm wondering whether these should be called unit:COUNT-PER-SEC-M2, etc. Any thoughts?

dr-shorthair:
Yes.

PR #126, opened 2020-05-23
Updated definition of RF-Power

alexrobin:


alexrobin:
You're right that I the "exactMatch" is incorrect.
However I was mostly trying to fix the definition of RF-Power that seem to be broken. Are you ok with this PR if I remove the DBm reference?

alexrobin:
I updated the PR according to my last comment

PR #127, opened 2020-05-29
unit:M2-PER-HA - square metres per hectare

edmondchuc:


edmondchuc:
@steveraysteveray thanks heaps for the explanation. A lot of it initially went over my head but I tried my best to make the changes you requested. 

1. I've updated the conversionMultiplier. 
2. I've defined `quantitykind:AreaRatio` by copying `quantitykindPressureRatio` as a template. The only thing I changed was the label and the numerator and denominator to your suggested `dimension:A0E0L2I0M0H0T0D0`. On `unit:M2-PER-HA`, `quantitykind:AreaRatio` was added as the value for `qudt:hasContextualQuantityKind` and `quantitykind:DimensionlessRatio` was added for `qudt:hasQuantityKind`.

jhodgesatmb:
Can you tell us, since it is probably fresh in your mind, how we might improve our instructions to make it easier for the next person to submit this kind of unit? Thank you.

Jack a Hodges

edmondchuc:
> Can you tell us, since it is probably fresh in your mind, how we might improve our instructions to make it easier for the next person to submit this kind of unit? Thank you.

I am not sure if there is anything that can be improved. A lot of it went over my head but the guidance from @steveraysteveray really helped.

PR #139, opened 2020-06-11
Rh quantity quantitykind

ralphtq:


steveraysteveray:
Ralph, I added hasContextualQuantityKind to Unit and Quantity in this PR. OK?

steveraysteveray:
I'm also wrestling with the notion of ContextualQuantityKind as a class. What it really implies is the use of the numerator and denominator properties pointing to dimension vectors. This might be the case for most of the QUDT dimensionless quantity kinds, but also for potential user-defined dimensionless quantity kinds. That suggests that this might be some sort of aspect class, since there could be multiple inheritance (contextual or not, user-defined or not). What's the most elegant way of capturing this?

steveraysteveray:
This concern is resolved because UserQuantityKind will always point to a qudt:QuantityKind, and if that object is a qudt:ContextualQuantityKind, then it will carry the numerator and denominator pointers. So only qudt:ContextualQuantityKind needs to have those properties.

PR #145, opened 2020-06-17
A number of cleanups of quantity kinds and referring entities

steveraysteveray:


dr-shorthair:
`breadth` ?
Depth is a very tricky one, mind you. Furniture dimensions and water column would look at it very differently.

steveraysteveray:
Agreed. The best I can think of is that the description is only suggestive of some uses, but not definitive. If someone really has heartache, they can suggest further specializations that have a skos:broader of depth.

jhodgesatmb:
Sadly, what people refer to as length, width, depth, height, breadth, etc are often, as Simon mentions, context dependent. But they are all measurements of size along a Cartesian dimension.

Jack Hodges, Ph.D.
Arbor Studios

jhodgesatmb:
Is Length a QuantityKind then, and the variations use skos:narrower?

Jack Hodges, Ph.D.
Arbor Studios

steveraysteveray:
Pretty much. The variations would use skos:broader to point to Length, or some other variant. 
See the skos taxonomy for Energy. 

 
- Steve

jhodgesatmb:
Ok

Jack Hodges, Ph.D.
Arbor Studios

PR #150, opened 2020-06-19
fixed skos:prefLabel values for GM_F and KiloGM_F units (had '?' in t…

jhodgesatmb:
…hem)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -9329,7 +9329,7 @@ unit:KiloLB_F-PER-FT
   qudt:uneceCommonCode "F17" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "KiloLB_F PER FT" ;
-  skos:prefLabel "pound?force per foot" ;
+  skos:prefLabel "pound force per foot" ;
```
steveraysteveray, line 9332: thousand pound, or kilopound (which sounds strange).
One or the other, right?

PR #152, opened 2020-06-22
Fixing a variety of SHACL constraint errors

steveraysteveray:


steveraysteveray:
In reverse order:
1. Yes, I did find more. However, your PR is still not yet merged. I approved it last week but left it unmerged because I had added a comment for you to consider.

2. Yes, I definitely agree there is much cleaning up to do with quantity kinds. The names are all over the place. We have some rules written in our [wiki,](https://github.com/qudt/qudt-public-repo/wiki/QuantityKind-Vocabulary-Submission-Guidelines#qname-naming-rules) but they are violated often. This will be a multi-week effort if not longer.

Since you mentioned CoM, it is also wrong because it has a skos:broader of Mass, which is definitely not true. It is a location (which we don't have in the vocabulary), or a coordinate, or something like that.

We need more people involved.

jhodgesatmb:
I certainly agree with your last comment.

Jack Hodges, Ph.D.
Arbor Studios

vocab/constants/VOCAB_QUDT-CONSTANTS-v2.1.ttl
```diff
@@ -2252,10 +2252,11 @@ constant:PermeabilityOfVacuum
 .
 constant:PermittivityOfVacuum
   a qudt:PhysicalConstant ;
-  dcterms:description "\"Permittivity of Vacuum}, also known as the \\textit{electric constant}, is a constant whose value is \\SI{\\approx 6.854188e-12}{\\farad\\per\\metre\"."^^rdf:HTML ;
+  dcterms:description "\"Permittivity of Vacuum}, also known as the \\textit{electric constant}, is a constant whose value is \\SI{\\approx 6.854188e-12}{\\farad\\per\\metre. Sometimes also referred to as the Permittivity of Free Space.\""^^rdf:HTML ;
```
ralphtq, line 2255: The \\SI LaTeX needs to be changed because the SI library is not in use any more.

Needs to be:

terms:description \(\\textit{Permittivity of Vacuum}\), also known as the \(\\textit{electric constant}\) is a constant whose value is \(\\approx 6.854188e-12\\,\\farad\\,\\per\\,metre\). Sometimes also referred to as the Permittivity of Free Space.\""^^qudt:LatexString ;

steveraysteveray: I thought you were fixing all the latex issues!

vocab/constants/VOCAB_QUDT-CONSTANTS-v2.1.ttl
```diff
@@ -2569,6 +2570,7 @@ constant:ReducedPlanckConstant
   a qudt:PhysicalConstant ;
   dcterms:description "The \"Reduced Planck Constant}, or Dirac Constant, is used in applications where frequency is expressed in terms of radians per second (\\emph{angular frequency\") instead of cycles per second. In such cases a factor of \\(2\\pi\\) is absorbed into the constant."^^rdf:HTML ;
   qudt:applicableUnit unit:J-SEC ;
+  qudt:hasQuantityKind quantitykind:Action ;
```
ralphtq, line 2573: Similar LaTeX errors in description.

Change to:

dcterms:description "The \\(\\textit{Reduced Planck Constant}\\), or \\(\\textit{Dirac Constant}\\), is used in applications where frequency is expressed in terms of radians per second (\\(\\textit{angular frequency}\\)) instead of cycles per second. In such cases a factor of \\(\\(2\\pi\\) is absorbed into the constant."^^qudt:LatexString ;

steveraysteveray: Ralph, a couple of questions:
1. Does this look correct, or should it have something about \textit:
```
qudt:LatexString
  rdf:type rdfs:Datatype ;
  rdfs:comment "A type of string in which some characters may be wrapped with '\\(' and '\\) characters for LaTeX rendering." ;
  rdfs:isDefinedBy <http://qudt.org/2.1/schema/qudt> ;
  rdfs:label "Latex String" ;
  rdfs:subClassOf xsd:string ;
.
```

2. Something isn't right here, and the description doesn't display at all in the normal EDG editor pane:
image

vocab/constants/VOCAB_QUDT-CONSTANTS-v2.1.ttl
steveraysteveray: Made the change, but see my comment on your next comment.

steveraysteveray:
I'm hoping we can go ahead and merge this PR, and go back and keep working on the dcterms:description issue discussed here.

steveraysteveray:
Good catch, thanks!
I have now added the CM, which I believe is 1000. I did a little research on this one and rewrote the latex, which now looks like this:
image

I _think_ I got it right, but you never know...

vocab/constants/VOCAB_QUDT-CONSTANTS-v2.1.ttl
ralphtq: Agree that we defer the fix

vocab/constants/VOCAB_QUDT-CONSTANTS-v2.1.ttl
ralphtq: Agree to deferring the fix

vocab/constants/VOCAB_QUDT-CONSTANTS-v2.1.ttl
ralphtq: I'd like to do the merge then look at this afterwards.

PR #153, opened 2020-06-24
Added millimetres per day

edmondchuc:


jhodgesatmb:
Perhaps he hasn't read/seen the unit submission guidelines because that is a required property in the guidelines. How do we make sure that potential contributors see the submission guidelines?

edmondchuc:
@steveraysteveray no problem, that was my bad, I forgot to add it.

@jhodgesatmb I have read the submissions guidelines but just forgot to do it since I read it a couple of weeks ago.

Ok, I've added it. 

An improvement could be to add an automatic test with GitHub Actions to test the vocabs file against a SHACL shape so that users who submit pull requests get notified immediately when a required property is expected.

edmondchuc:
I just remembered that I didn't add the rdfs:isDefinedBy property for this pull request https://github.com/qudt/qudt-public-repo/pull/154. Another good reason to add automated testing since us humans may miss it from time to time.

PR #154, opened 2020-06-25
Added missing required property rdfs:isDefinedBy

edmondchuc:


steveraysteveray:
I believe the conflict relates to the two units MilliM-PER-DAY and deciS-PER-M. Both units look good in the master branch, so I am closing this PR.
The issue came up (I believe) because the branch head was very old (over 91 commits made since then).

PR #181, opened 2020-07-22
Added delta property to Quantity, and cleaned up collections.

steveraysteveray:


jhodgesatmb:
I thought we were going to rename the collection file to not have ui.ttl in the name.

PR #189, opened 2020-07-23
 Fixed all the 0,0x-fold comments. Changed to 0.0x-fold

steveraysteveray:


jhodgesatmb:
I fixed those many times and somehow they got unfixed. Thanks for doing this. Perhaps when you do it it will stay fixed.

PR #194, opened 2020-07-29
Changes done for EDG co-existence

ralphtq:


schema/shacl/SHACL-SCHEMA-SUPPLEMENT_QUDT-v2.1.4.ttl
```diff
@@ -1,5 +1,5 @@
 # baseURI: http://qudt.org/2.1/shacl/schema/overlay/qudt
```
steveraysteveray, line 1: Shouldn't the baseURI be http://qudt.org/2.1/schema/shacl/overlay/qudt ?

Issue #197, opened 2020-08-06
inconsistency between all-units HTML page and corresponding Turtle file


Toni100:
There are units which appear on http://qudt.org/doc/2020/03/DOC_VOCAB-UNITS-ALL-v2.1.html but are missing from the Turtle file qudt.org/2.1/vocab/unit .
For instance: unit:LB_M
Expected: The website and the corresponding Turtle file should contain exactly the same content.

steveraysteveray:
Ah, I see the problem. You are referring to the March, 2020 version of the HTML file (indicated in the URL). If you go to qudt.org and follow the link to the 2.1 catalog, you will (currently) be directed to the July, 2020 version of the same file. I think you will find that the latest version of the HTML file will always be consistent with the resolved link qudt.org/2.1/vocab/unit. I will be regenerating all the html files soon, in which case you will see the August, 2020 versions.

Regarding the unit:LB_M unit in particular, we decided in early July that it was possibly confusing to people to require either a _F or _M for all unit:LB occurrences. So we made the default (without any "_") to mean "mass", and only have _F for the "force" variant. This is documented in Rule 2 [here](https://github.com/qudt/qudt-public-repo/wiki/Unit-Vocabulary-Submission-Guidelines#qname-naming-rules).

Toni100:
Could there be a generic URL which does not contain any date or minor version information, but always shows the content of the latest published version? This should not be a redirect: In that way, a user, coming back now and then to the bookmarked URL will always see the most up to date content.
This document could also contain longer URLs containing date or minor version for reference, should that be necessary for some use case.
This is how the W3C handles their standards: For instance, [1] shows the latest, final version, but years ago used to show the various developmental versions. Note the "This Version" section, which states the long, versioned, stable URL of the current document.

[1] https://www.w3.org/TR/rdf11-concepts/

jhodgesatmb:
In a way this goes back to an earlier discussion about whether to use versionIRI for versioning and keep the baseURI constant. That said, I was under the impression that the HTML always points to the latest version of the vocabularies, and that you have to specifically ask for another version if you want it. Maybe this is just a referencing question...

Jack Hodges, Ph.D.
Arbor Studios

steveraysteveray:
[...]

PR #200, opened 2020-08-14
Republished after versionless html capability

steveraysteveray:


dr-shorthair:
fixes #197?

steveraysteveray:
Related, but not really. Every time I regenerate the html pages, the metadata of the ttl files is updated. So in this case, while it's true that I regenerated the html pages to fix #197, this kind of PR will keep popping up in the future as well.

PR #218, opened 2020-09-01
Alternative labels and symbols for parts-per-thousand

dr-shorthair:


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -15604,10 +15604,14 @@ unit:PPTH
...
   qudt:hasQuantityKind quantitykind:DimensionlessRatio ;
   qudt:informativeReference "http://aurora.regenstrief.org/~ucum/ucum.html#section-Derived-Unit-Atoms"^^xsd:anyURI ;
+  qudt:abbreviation "‰" ;
+  qudt:literal "‰" ;
+  qudt:symbol "‰" ;
   qudt:symbol "ppth" ;
   qudt:ucumCode "[ppth]" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Parts per thousand" ;
+  rdfs:label "per mil" ;
```
steveraysteveray, line 15614: We aren't allowing multiple rdfs:label values. It makes the EDG displays unpredictable regarding which one it will pick. Other than that, looks fine.

dr-shorthair: Is `skos:altLabel` an option?

steveraysteveray: We have used skos:altLabel for a handful of quantity kinds, although I notice it's not in the schema. I'm fine with adding it.

Other voices?

PR #225, opened 2020-09-04
Added  UCUM license text, 

dr-shorthair:
as required by Clause 7 of https://ucum.org/trac/wiki/TermsOfUse

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -29,6 +29,11 @@
   owl:imports <http://qudt.org/2.1/vocab/quantitykind> ;
   owl:imports <http://qudt.org/2.1/vocab/sou> ;
   owl:versionIRI <http://qudt.org/2.1/vocab/unit> ;
+  dcterms:rights """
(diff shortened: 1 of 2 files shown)


steveraysteveray (commented):
I have no objections to this. Could others at QUDT.org weigh in?

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -29,6 +29,11 @@
   owl:imports <http://qudt.org/2.1/vocab/quantitykind> ;
   owl:imports <http://qudt.org/2.1/vocab/sou> ;
   owl:versionIRI <http://qudt.org/2.1/vocab/unit> ;
+  dcterms:rights """
```
steveraysteveray, line 32: This would be better placed in the metadata instance: vaem:GMD_QUDT-UNITS-ALL later in this file.

Also, could you do the same in the QUDT Schema (http://qudt.org/2.1/schema/qudt), again in the metadata instance?

Thanks.

steveraysteveray (changes requested):

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
dr-shorthair: Done.

steveraysteveray (approved):

PR #233, opened 2020-09-17
Simplifying all temperature units for now, to just point to quantitykind:Temperature

steveraysteveray:
Later, we should resolve the distinction from ThermodynamicTemperature, CelsiusTemperature...


jhodgesatmb:
Are we going to do this for all units? It seems too narrow to only do temperature.

Jack Hodges, Ph.D.
Arbor Studios

steveraysteveray:
I addressed Temperature because there is some ambiguity in how we should relate Temperature, CelsiusTemperature and ThermodynamicTemperature, using skos:broader. This was causing a problem with defining constraints in my project.

We DO have well-defined skos relationships for some of the other quantity kinds, and that's a good thing.

vocab/prefixes/VOCAB_QUDT-PREFIXES-v2.1.ttl
```diff
@@ -338,6 +338,7 @@ prefix:Zetta
   qudt:ucumCode "Z" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/prefix> ;
   rdfs:label "Zetta" ;
+  rdfs:label "Zzzettttaaaa"@en-us ;
```
ralphtq, line 341: ???

ralphtq (approved):

PR #241, opened 2020-09-28
UCUM codes for the derived units

dr-shorthair:
Tedious work, trying to do a bit at a time. 
Draft PR for the moment as I've only got up to the megas and millis so far.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -50,7 +50,7 @@ Note that SI supports only the use of symbols and deprecates the use of any abbr
   qudt:informativeReference "http://en.wikipedia.org/wiki/Ampere?oldid=494026699"^^xsd:anyURI ;
   qudt:omUnit <http://www.ontology-of-units-of-measure.org/resource/om-2/ampere> ;
   qudt:symbol "A" ;
-  qudt:ucumCode "A" ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (commented):
First, thanks so much for doing this thankless work!
(Although, I suppose by thanking you it is no longer thankless :)

I notice you have some values typed ^^qudt:UCUMcs, and others are not typed. Shouldn't they all have that type? If so, an easy SPARQL fix.

dr-shorthair:
Yes, they should all be `^^qudt:UCUMcs`. It gets set automatically by TBCME when entering values using the `Instances` tab, but not if you use the TBCME Form. I was planning to fix them all at the end with a SPARQL query. 

FWIW this validation tool is quite nice, and provides a consistent textual label for any valid UCUM code - https://ucum.nlm.nih.gov/ucum-lhc/demo.html

steveraysteveray (approved):

dr-shorthair:
Just discovered this is incomplete, because I had max # instances=1000 in TopBraid, while there are >2000 units. 
More drudgery.

steveraysteveray:
Your efforts are greatly appreciated!!!

dr-shorthair:
Well I have a great drive towards harmonization, and if we could get QUDT+UCUM playing nice together, then I think we win. 

UCUM has a huge penetration in biomedical. I understand it is built in to FHIR which is the main game now. It is also very rational and generally highly consistent. A few gaps, but not many. Best would be to do all this algorithmically of course, but that's beyond my skillset.

PR #246, opened 2020-10-01
More UCUM codes

dr-shorthair:
I missed a load first time through due to truncation of the list in TopBraid ;-) 
I've also normalized them all to .x-1 instead of /x form, to match Green Book conventions.


dr-shorthair:
(and picked up a few errors along the way - I'm sure there will be a few remaining, but this is better now.)

steveraysteveray (approved):

PR #255, opened 2020-10-09
Moved currency units into separate graph

dr-shorthair:
N.B. Metadata in currency graph needs correcting - I just cloned and modified a bit but don't know the expected pattern

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -306,28 +306,6 @@ unit:AC-FT
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Acre Foot"@en ;
 .
-unit:AFN
-  a qudt:CurrencyUnit ;
-  a qudt:Unit ;
-  dcterms:description "Afghanistan"^^rdf:HTML ;
-  qudt:currencyExponent 2 ;
-  qudt:dbpediaMatch "http://dbpedia.org/resource/Afghani"^^xsd:anyURI ;
-  qudt:hasQuantityKind quantitykind:Currency ;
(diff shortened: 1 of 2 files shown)


steveraysteveray (commented):
Looks pretty good! I may need to tweak a couple of tiny things in the GMD graph meta-data part, because my publication scripts rely heavily on 8 of the values there. However, the best way to find that out is to just try it after I update the scripts and catalog entries to reflect this separate graph.   For these reasons, I'm going to leave this as a branch and not merge it until everything is working correctly.

steveraysteveray:
Simon,
I made a bunch of changes to the new currency graph in your fork, including adding the currency: prefix.
However, after additional discussion, the QUDT Board has decided we should hold off on splitting off this set of units as a distinct graph, primarily for the reasons we went to all the trouble of consolidating our many unit graphs into one, last year. So we'd like to put this PR on the back burner for now. Let's keep the branch around, though. We can continue to discuss in the Technical Advisory Board.

dr-shorthair:
Understood. However, I do think this is a special case. (a) it is not used in measurements - there is no sensor to measure an amount of a currency - and (b) the conversion factors are time-dependent.

dr-shorthair:
Did you intend to close this? Labeling it low priority makes sense, but closing without merging looks like rejection.

steveraysteveray:
Not at all my intention. In fact we should consider a couple of steps anyway:
1. You could resync your fork and resolve the merge conflicts for this branch
2. We could create a branch in the qudt repo and merge to that, so that this PR doesn't just exist in your fork

Up to you how much time you want to devote!

dr-shorthair:
I'd be much happier with a branch in the QUDT repo, instead of having to keep syncing my fork. But I don't have push permission so can't work on that myself.

steveraysteveray:
I just created a branch called "currency" in qudt. Once you resync and resolve conflicts, see if you can issue a PR to that branch rather than to master.

PR #257, opened 2020-10-13
Correction of Dimension Vector for Moment of Force

cpauloh:


steveraysteveray (changes requested):
Good catch. Could you go ahead and also delete
  qudt:baseUnitDimensions "kg \\cdot m^3/s" ;

steveraysteveray (approved):

PR #258, opened 2020-10-13
Coulomb is not an applicable unit for Electric Flux

cpauloh:


steveraysteveray:
Agreed. Once we reach consensus on how we will generate the applicableUnit values, we can clean them all up.

steveraysteveray (approved):

Issue #243, opened 2020-10-01
Duplicate units?


dr-shorthair:
unit:CAL_TH-PER-G
unit:CAL_TH-PER-GM

steveraysteveray:
Good catch. We are using unit:GM for gram, and unit:G for "gravity", so the first one is not just a duplicate, it is an error. Where possible, we intend to align with UCUM for the choice of qname. In this case, one is g and the other is [g]. I think we made the right call here in using G for gravity and GM for gram.

You (or I) could remove the unit:CAL_TH-PER-G
@dr-shorthair, grab the assignment if you intend to do it.

steveraysteveray:
Done.

PR #271, opened 2020-10-30
Issue#243

dr-shorthair:
Potential fix for https://github.com/qudt/qudt-public-repo/issues/243 
I've marked the individual 'deprecated'. 
Alternatively, it could be updated to match the ID (Gravity not Grams) - though that is not a very useful unit.


steveraysteveray (commented):
The changes look valid to me, but I'm not well-schooled in datatypes and how QUDT uses them. Therefore, I'm hoping Ralph can also take a look.

steveraysteveray (approved):

Issue #268, opened 2020-10-27
NVS P06 units that are missing from QUDT


dr-shorthair:
This list attached here https://github.com/qudt/qudt-public-repo/issues/268#issuecomment-721424132 shows items from the [NERC Vocabulary Service P06 vocabulary](vocab.nerc.ac.uk/collection/P06/current/) which do not have matches in QUDT, but could do since the elements are already there: 

Also see https://github.com/nvs-vocabs/P06/issues/18 

@gwemon

dr-shorthair:
Here they are with Dimension vectors and conversion-multipliers

 QUDT item requested | Dimension vector | Numerator | Denominator | Multiplier | UCUM code 
 --- |  --- |  --- |  --- |  --- | --- 
 unit:BQ-SEC-PER-M3 | A0E0L-3I0M0H0T0D0 |  |  | 1 | Bq.s.m-3 
 unit:DeciBAR-PER-YR | A0E0L-1I0M1H0T-3D0 |  |  | 1.00E-18 | dbar.a-1 
 unit:DEG_C-PER-YR | A0E0L0I0M0H1T-1D0 |  |  | 1 | Cel.a-1 
 unit:DEG-PER-M | A0E0L-1I0M0H0T0D1 |  |  | 1 | deg.m-1 
 unit:DEG_C-KiloGM-PER-M2 | A0E0L-2I0M1H1T0D0 |  |  | 1 | Cel.kg.m-2 
 unit:K-PER-M | A0E0L-1I0M0H1T0D0 |  |  | 1 | K.m-1 
 unit:FemtoGM-PER-KiloGM | A0E0L0I0M0H0T0D1 | A0E0L0I0M1H0T0D0 | A0E0L0I0M1H0T0D0 | 1.00E-18 | fg.kg-1 
 unit:FemtoGM-PER-L | A0E0L-3I0M1H0T0D0 |  |  | 1.00E-15 | fg.L-1 
 unit:FemtoMOL-PER-KiloGM | A1E0L0I0M-1H0T0D0 |  |  | 1.00E-15 | fmol.kg-1 
 unit:GM-PER-CentiM2-YR | A0E0L-2I0M1H0T-1D0 |  |  | 3.17E-07 | g.cm-2.a-1 
 unit:HectoPA-PER-HR | A0E0L-1I0M1H0T-3D0 |  |  | 0.027777778 | hPa.h-1 
[...]

PR #292, opened 2020-11-23
P06 additions from spreadsheet (converted to TTL using SKOS-Play!)

dr-shorthair:


steveraysteveray:
Quick review of your units.ttl file. This is much easier - will definitely think about making the SKOS-Play! spreadsheet template into a submission template.

Rather than my cloning your fork and submitting a branch back to you, would you consider just making the following edits?

Add the following triple to all new units:

Change all labels starting with "per ..." to "Reciprocal ..." (12 instances)

Replacing all strings of "by weight" in dct:description to "by mass" (4 instances)

Curious why you decided on the qname unit:PH_ACIDITY instead of just unit:PH? I recall from high school that there is also POH for the OH radical, but not sure if P<something> occurs elsewhere?
Also, on PH_ACIDITY, wouldn't the dimension be A1L-3? (Amount of substance per volume, right?) Admittedly not linear, but we can capture the dimension.

Typo:
unit:BQ-SEC-PER-M3 a qudt:Unit;
  dct:description "TBC"@en
to
unit:BQ-SEC-PER-M3 a qudt:Unit;
  dct:description "TBD"@en

In rdfs:label, change all "xyz squared" to "square xyz" (11 instances)

steveraysteveray:
Thinking about PH, since it is the negative logarithm, maybe the dimension is the reciprocal...  A-1L3?

dr-shorthair:
I've processed all the changes requested, and added a comment in the entry for PH.

dr-shorthair:
@gwemon @nicholascar We are most of the way now. Just a whole lot of missing quantity-kind links from unit definitions.

steveraysteveray (approved):
We can discuss the dimensionality of PH going forward...

dr-shorthair:
Fixes #268

PR #293, opened 2020-12-01
Fix two incorrect multipliers

dr-shorthair:


steveraysteveray (approved):
Nice catches. Had to think twice about the second one, then remembered radians.

PR #295, opened 2020-12-01
header comment incorrect

cpauloh:
The ontology import declaration is correct, but the header comment doesn't mirror the import declaration.

This is a nuisance when uploading qudt as a project in EDG.  The header comment `# imports:` appears to be used by EDG determine required imports and warns the user about missing imports.


steveraysteveray:
All of the vocab graphs that do NOT appear in our [catalog](http://www.qudt.org/2.1/catalog/qudt-catalog.html) have not been scrubbed or analyzed since we moved to 2.1. Therefore, they are really placeholders. I will admit that I haven't tried loading them into an EDG collection. But are you needing either the unicode graph or the scales graph? 

Again, with our limited resources, we're taking these things one at a time as we have the time.

cpauloh:
This is just a change to a comment to reduce the number of EDG errors or warnings when QUDT is uploaded as a project.  Since we're trying to use QUDT for enterprise applications, this removes a need to track and change it every time a new release is loaded.

steveraysteveray (approved):

Issue #303, opened 2020-12-07
Wrong conversion factors


jmkeil:
Using a [comparison of unit ontologies with ABECTO](https://github.com/fusion-jena/abecto-unit-ontology-comparison) I became aware that the following units have wrong conversion factors:
* [ ] http://qudt.org/vocab/unit/KiloA: `1.03e-3`
* [ ] http://qudt.org/vocab/unit/MegaL: `1e+06`
* [ ] http://qudt.org/vocab/unit/MilliM_HG: `86014.2806`
* [ ] http://qudt.org/vocab/unit/C_Stat: `3.33564E-10`
* [ ] http://qudt.org/vocab/unit/FARAD_Stat: `1.113E-12` (strongly rounded)
* [ ] http://qudt.org/vocab/unit/FT-PDL: `0.042140124`
* [ ] http://qudt.org/vocab/unit/TON_FG: `3517` (strongly rounded)
* [ ] http://qudt.org/vocab/unit/HART: `1.0` (to http://qudt.org/vocab/unit/NAT)
* [ ] http://qudt.org/vocab/unit/GigaBYTE: `1073741824` (to http://qudt.org/vocab/unit/NAT)
* [ ] http://qudt.org/vocab/unit/KiloBYTE: `1024` (to http://qudt.org/vocab/unit/NAT)
* [ ] http://qudt.org/vocab/unit/MegaBYTE: `1048576` (to http://qudt.org/vocab/unit/NAT)
* [ ] http://qudt.org/vocab/unit/PebiBYTE: `1125899906842624` (to http://qudt.org/vocab/unit/NAT)
* [ ] http://qudt.org/vocab/unit/PetaBYTE: `1.0e15` (to http://qudt.org/vocab/unit/NAT)
* [ ] http://qudt.org/vocab/unit/TeraBYTE: `1048576` (to http://qudt.org/vocab/unit/NAT)

Regarding the units of information, things might become easier if Bit or Byte would be considered as the reference unit for conversions. Same applies for units derived from the units of information. There is also inconsistent handling of the prefix issue.

Further there might be wrong values at the following units, but I am not able to decide upon without further investigation:
* [ ] http://sweetontology.net/reprSciUnits/year with `3.1556926E7` vs. `3.1536e7` by http://www.ontology-of-units-of-measure.org/resource/om-2/year vs. `31557600` by http://qudt.org/vocab/unit/YR (probably different definitions have been used, but QUDT maps to this OM instance)
[...]

PR #310, opened 2020-12-09
Fix a few conversion factors - see #303

dr-shorthair:
Can't find reliable value for Tons of Refrigeration, or the NAT-based units

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -17586,7 +17586,7 @@ unit:PER-GM
 .
 unit:PER-GigaEV2
   a qudt:Unit ;
-  dcterms:description "Per Square Giga Electron Volt Unit is a a demominator unit with dimensions \\(/GeV^2\\)."^^qudt:LatexString ;
+  dcterms:description "Per Square Giga Electron Volt Unit is a a denominator unit with dimensions \\(/GeV^2\\)."^^qudt:LatexString ;
   qudt:conversionMultiplier 38956440500000000000.0 ;
   qudt:expression "\\(/GeV^2\\)"^^qudt:LatexString ;
   qudt:hasDimensionVector qkdv:A0E0L-4I0M-2H0T4D0 ;
@@ -17641,7 +17641,7 @@ unit:PER-J-M3
 .
 unit:PER-K
   a qudt:Unit ;
-  dcterms:description "Per Kelvin Unit is a demominator unit with dimensions \\(/k\\)."^^qudt:LatexString ;
+  dcterms:description "Per Kelvin Unit is a denominator unit with dimensions \\(/k\\)."^^qudt:LatexString ;
   qudt:conversionMultiplier 1.0 ;
   qudt:expression "\\(/K\\)"^^qudt:LatexString ;
   qudt:hasQuantityKind quantitykind:CubicExpansionCoefficient ;
@@ -17679,7 +17679,7 @@ unit:PER-L
 .
 unit:PER-M
   a qudt:Unit ;
-  dcterms:description "Per Meter Unit is a a demominator unit with dimensions \\(/m\\)."^^qudt:LatexString ;
+  dcterms:description "Per Meter Unit is a a denominator unit with dimensions \\(/m\\)."^^qudt:LatexString ;
   qudt:conversionMultiplier 1.0 ;
   qudt:derivedCoherentUnitOfSystem <http://qudt.org/vocab/sou/SOU_SI> ;
   qudt:expression "\\(per-meter\\)"^^qudt:LatexString ;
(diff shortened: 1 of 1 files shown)


steveraysteveray:
Our current entry for unit:TON_FG has two references. Not good enough?

dr-shorthair:
The Wikipedia entry does not give a good conversion factor. 
OM is a bit better. I'll use that one.

dr-shorthair:
Fixes #303

steveraysteveray (changes requested):
Simon,
Thanks so much for getting into this and fixing various conversion factors. In reviewing your work, some pre-existing errors have come to light. Specifically, QUDT has interpreted the combination of the Kilo, Mega, Peta, etc. prefixes with binary units (Bit, Byte) to mean powers of 2, such as Kilo meaning 1024, etc. Assuming we continue with this approach, there seem to be some inconsistencies, such as the conversionMultiplier for PetaByte being exactly 1000 times larger than the one for TeraBYTE.
Would you mind reviewing and fixing the various BIT and BYTE units to ensure they are all consistently powers of 2 rather than powers of 10?

Meanwhile, we can consider whether we want to augment the vocabulary with something like KiloBYTE_decimal for the 1000x version that many people also use.

dr-shorthair:
Yes - I struggled a bit with these. 
First: do we want to use NAT as the reference unit for information?

steveraysteveray:
Yes, the documentation indicates that the SI unit is indeed the NAT, so that looks good.

jhodgesatmb (approved):

dr-shorthair:
Now I've added the 'binary' equivalents all the way up to 'exbi'. 
However, I've followed https://en.wikipedia.org/wiki/Byte#Multiple-byte_units and used the binary multiplier 1024, 1024^3, 1024^6 for kilo, mega and giga. 

Also see https://github.com/qudt/qudt-public-repo/issues/303#issuecomment-749894875

Should be good to go now.

dr-shorthair:
unit:ExaBYTE 5545177444479562475.3378569716654
unit:ExbiBYTE 6393154322601327829.8943153498712
unit:GibiBYTE 5954088943.6391441429912255610071
unit:GigaBYTE 5954088943.6391441429912255610071
unit:KibiBYTE 5678.2617031470719747459655389854
unit:KiloBYTE 5678.2617031470719747459655389854
unit:MebiBYTE 5814539.984022601702139868711921
unit:MegaBYTE 5814539.984022601702139868711921
unit:PebiBYTE 6243314768165359.2088811673338586
unit:PetaBYTE 5545177444479562.4753378569716654
unit:TebiBYTE 6096987078286.4836024230149744713
unit:TeraBYTE 5545177444479.5624753378569716654

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -5544,6 +5544,21 @@ unit:EuropeanUnitOfAccount9
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "European Unit of Account 9 (E.U.A.-9) (Bonds market unit)"@en ;
 .
+unit:ExaBYTE
+  a qudt:Unit ;
+  dcterms:description "The exabyte is a multiple of the unit byte for digital information. The prefix exa means 10^18 in the International System of Units (SI), however, in information quantities kilo=1024 so ExaByte is 1024^6 Bytes."^^rdf:HTML ;
```
steveraysteveray, line 5549: This description needs to be adjusted to state that ExaBYTE is 10^18 Bytes and not 1024^6 Bytes.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -9395,7 +9450,8 @@ unit:KiloBYTE-PER-SEC
   a qudt:DerivedUnit ;
   a qudt:Unit ;
   dcterms:description "A kilobyte per second (kByte/s) is a unit of data transfer rate equal to 8,000 bits per second, or 1,000 bytes per second, or 8 kilobits per second."^^rdf:HTML ;
```
steveraysteveray, line 9452: Need to change to say 1024 bytes per second or 8096 bits per second.

steveraysteveray (changes requested):
See just the few in-line comments that need fixing, related to the dcterms:description fields.
Almost there!

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
dr-shorthair: Changed per this review

steveraysteveray (approved):

steveraysteveray:
Whoops. More pesky merge conflicts!

dr-shorthair:
Conflicts resolved

PR #311, opened 2020-12-09
Convert doubles and floats to decimals

### LINKED ISSUE #307: Use of xsd:decimal instead of xsd:double for conversion factors
Using a [comparison of unit ontologies with ABECTO](https://github.com/fusion-jena/abecto-unit-ontology-comparison) I became aware that QUDT uses `xsd:double` to represent conversion factors and offsets. I think, instead `xsd:decimal` should be used, as binary floating point datatypes are not able to exactly represent many of the conversion values. Especially 0.1, 0.01, 0.001, … can not be represented. Only the lexical representation seems to be exact. However, the lexical to value mapping will map them to a slightly different value. This runs the risk of numerical problems in applications. I outlined this in detail in [arXiv:2011.08077](https://arxiv.org/abs/2011.08077).

dr-shorthair:
Fixes #307

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -48,7 +48,6 @@ Note that SI supports only the use of symbols and deprecates the use of any abbr
   qudt:allowedUnitOfSystem <http://qudt.org/vocab/sou/SOU_CGS-GAUSS> ;
   qudt:allowedUnitOfSystem <http://qudt.org/vocab/sou/SOU_PLANCK> ;
   qudt:conversionMultiplier 1.0 ;
-  qudt:conversionOffset 0.0 ;
   qudt:dbpediaMatch "http://dbpedia.org/resource/Ampere"^^xsd:anyURI ;
   qudt:hasDimensionVector qkdv:A0E1L0I0M0H0T0D0 ;
   qudt:hasQuantityKind quantitykind:CurrentLinkage ;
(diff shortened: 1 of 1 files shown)


dr-shorthair:
Used the following SPARQL

INSERT { ?u qudt:conversionMultiplier ?md }
WHERE {
	?u a qudt:Unit ;
		qudt:conversionMultiplier ?m .
	BIND( DATATYPE(?m) as ?mt )
	FILTER( ?mt != xsd:decimal )
	BIND( xsd:decimal(?m) as ?md )
}

DELETE { ?u qudt:conversionMultiplier ?m }
WHERE {
	?u a qudt:Unit ;
		qudt:conversionMultiplier ?m .
	BIND( DATATYPE(?m) as ?mt )
	FILTER( ?mt != xsd:decimal )
}

INSERT { ?u qudt:conversionOffset ?od }
WHERE {
	?u a qudt:Unit ;
		qudt:conversionOffset ?o .
	BIND( DATATYPE(?o) as ?ot )
	FILTER( ?ot != xsd:decimal )
	BIND( xsd:decimal(?o) as ?od )
}

DELETE { ?u qudt:conversionOffset ?o }
WHERE {
	?u a qudt:Unit ;
		qudt:conversionOffset ?o .
	BIND( DATATYPE(?o) as ?ot )
	FILTER( ?ot != xsd:decimal )
}

Also - remove all zero offets

DELETE { ?u qudt:conversionOffset ?o }
WHERE {
	?u a qudt:Unit ;
		qudt:conversionOffset ?o .
	FILTER( ?o = 0.0 )
}

nicholascar:
> remove all zero offets

Is this because it is inferred to be zero if not present by calculation queries?

dr-shorthair:
Offsets apply in very few cases - and only on interval scales (not ratio scales). 
As they are the exceptional case, they need not be present in general. 
(AFAICT there are only 3 valid offsets in the current catalogue.)

steveraysteveray (approved):
OK.
I hope the multiplier for PlanckDensity doesn't break anything!

Issue #272, opened 2020-11-01
Relation from Unit to QuantityKind should not be 'mandatory`


dr-shorthair:
In the guidelines for submission of new units `qudt:hasQuantityKind` is mandatory. 
The QUDT Schema (ontology) gives it a max count of '1'. 

I don't think this is such a good idea. 

1. some units are related to more than one QuantityKind - particularly when we consider specializations of those
2. the inverse relationship (`qudt:applicableUnit`), from a `qudt:QuantityKind` to a `qudt:Unit` is more relevant. 

I'd rather see `qudt:hasDimensionVector` as mandatory, and `qudt:hasQuantityKind` as optional and repeatable.

steveraysteveray:
I'm totally with you on item 1, because of the new property qudt:appropriateUnit (which is a magic property in the OWL version, and a SHACL shape in the SHACL version, soon to be merged hopefully). This new approach requires some units to relate to more than one QuantityKind. However, I'm not sure when we would want qudt:hasQuantityKind to be absent (i.e. optional). Can you think of a use case for that?

qudt:appropriateUnit will eventually replace qudt:applicableUnit, as it is inferred rather than manually maintained, and thus much more manageable. Because of this, your item 2 really relates to qudt:appropriateUnit. 

The algorithm for inferring qudt:appropriateUnit is illustrated here (the legend text should say "appropriateUnit" rather than "applicableUnit"):
![image](https://user-images.githubusercontent.com/1130189/97813290-fc82a280-1c3b-11eb-837c-7fa05df02e4a.png)

Don't get too hung up on the actual example scenario, which is not reflected precisely in the vocabulary. I was just trying to illustrate that sometimes a more specialized quantity kind could have a subset, or even a different set of appropriateUnit values (such as in the case of reactive power).
Note that it works by examining the qudt:hasQuantityKind values coming from instances of Unit. This means we need to go in and clean up some of the qudt:hasQuantityKind values so that we get the desired results for qudt:appropriateUnit.
FYI, the "almost finished" version of the algorithm is currently running on our EDG server, so you can check it out and compare to the qudt:applicableUnit values.

dr-shorthair:
I'll explain the scenario which led to this: 

As you are aware, I'm looking at the NVS units vocabulary 'P06' -http://vocab.nerc.ac.uk/collection/P06/current/ 
[...]

PR #320, opened 2020-12-17
Loop3d units

nicholascar:
Units required by the Loop3D project.

These units, plus a few more, will appear in a Community Profile about to be PRed across also.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -8781,7 +8781,7 @@ unit:KiloGM-PER-M3
   a qudt:DerivedUnit ;
   a qudt:Unit ;
   qudt:plainTextDescription "Kilogram per cubic metre is an SI derived unit of density, defined by mass in kilograms divided by volume in cubic metres. The official SI symbolic abbreviation is \\(kg \\cdot m^{-3}\\), or equivalently either \\(kg/m^3\\)."^^qudt:LatexString ;
-  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionMultiplier 1.0 ;
   qudt:conversionOffset 0.0 ;
   qudt:expression "\\(kg/m^3\\)"^^qudt:LatexString ;
   qudt:hasDimensionVector qkdv:A0E0L-3I0M1H0T0D0 ;
@@ -18532,7 +18532,7 @@ unit:SUSCEPTIBILITY_ELEC
   a qudt:DerivedUnit ;
   a qudt:Unit ;
   qudt:plainTextDescription "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." ;
-  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionMultiplier 1.0 ;
   qudt:conversionOffset 0.0 ;
   qudt:latexDefinition "\\chi_{\\text{e}} = \\frac{{\\mathbf P}}{\\varepsilon_0{\\mathbf E}}"^^qudt:LatexString ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
@@ -18544,7 +18544,7 @@ unit:SUSCEPTIBILITY_MAG
   a qudt:DerivedUnit ;
   a qudt:Unit ;
   qudt:plainTextDescription "Magnetic susceptibility is a dimensionless proportionality constant that indicates the degree of magnetization of a material in response to an applied magnetic field. Here M = chi * H. Where M is the magnetization of the material (the magnetic dipole moment per unit volume), measured in amperes per meter, and H is the magnetic field strength, also measured in amperes per meter. Chi is therefore a dimensionless quantity." ;
-  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionMultiplier 1.0 ;
   qudt:conversionOffset 0.0 ;
   qudt:latexDefinition "\\chi_\\text{v} = \\frac{\\mathbf{M}}{\\mathbf{H}}"^^qudt:LatexString ; 
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
@@ -20192,7 +20192,7 @@ unit:MegaYR
     a qudt:Unit ;
     a qudt:DerivedUnit ;
     qudt:plainTextDescription "1,000,000-fold of the derived unit year." ;
-    qudt:conversionMultiplier 1e+06 ;
+    qudt:conversionMultiplier 1000000.0 ;
     qudt:conversionOffset 0.0 ;
     qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T1D0 ;
     qudt:hasQuantityKind quantitykind:Time ;
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -18530,10 +18530,10 @@ unit:SurinamDollar
 unit:SUSCEPTIBILITY_ELEC
   a qudt:DerivedUnit ;
   a qudt:Unit ;
-  qudt:plainTextDescription "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." @en ;
+  qudt:plainTextDescription "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." ;
   qudt:conversionMultiplier 1.0e0 ;
   qudt:conversionOffset "0.0"^^xsd:double ;
-  qudt:latexDefinition "\chi_{\text{e}}=\frac{{\mathbf P}}{\varepsilon_0{\mathbf E}}"^^qudt:LatexString ;
+  qudt:latexDefinition "\\chi_{\\text{e}} = \\frac{{\\mathbf P}}{\\varepsilon_0{\\mathbf E}}"^^qudt:LatexString ;
   # qudt:hasQuantityKind quantitykind:ElectricFluxDensity ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Electric Susceptibility" ;
@@ -18541,10 +18541,10 @@ unit:SUSCEPTIBILITY_ELEC
 unit:SUSCEPTIBILITY_MAG
   a qudt:DerivedUnit ;
   a qudt:Unit ;
-  qudt:plainTextDescription "Magnetic susceptibility is a dimensionless proportionality constant that indicates the degree of magnetization of a material in response to an applied magnetic field. Here M = chi * H. Where M is the magnetization of the material (the magnetic dipole moment per unit volume), measured in amperes per meter, and H is the magnetic field strength, also measured in amperes per meter. Chi is therefore a dimensionless quantity." @en ;
+  qudt:plainTextDescription "Magnetic susceptibility is a dimensionless proportionality constant that indicates the degree of magnetization of a material in response to an applied magnetic field. Here M = chi * H. Where M is the magnetization of the material (the magnetic dipole moment per unit volume), measured in amperes per meter, and H is the magnetic field strength, also measured in amperes per meter. Chi is therefore a dimensionless quantity." ;
   qudt:conversionMultiplier 1.0e0 ;
   qudt:conversionOffset "0.0"^^xsd:double ;
-  qudt:latexDefinition "\chi_\text{v} =\frac{\mathbf{M}}{\mathbf{H}}"^^qudt:LatexString ; 
+  qudt:latexDefinition "\\chi_\\text{v} = \\frac{\\mathbf{M}}{\\mathbf{H}}"^^qudt:LatexString ; 
   # qudt:hasQuantityKind quantitykind:MagneticFlux ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Magnetic Susceptibility" ;
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -18513,6 +18513,30 @@ unit:SurinamDollar
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Surinam Dollar" ;
 .
+unit:SUSCEPTIBILITY_ELEC
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  dcterms:description "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." @en ;
+  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionOffset "0.0"^^xsd:double ;
+  qudt:latexDefinition "\chi_{\text{e}}=\frac{{\mathbf P}}{\varepsilon_0{\mathbf E}}"^^qudt:LatexString ;
+  # qudt:expression "\varepsilon_0\chi_{\text{e}}{\mathbf E}"^^qudt:LatexString ;
+  # qudt:hasQuantityKind quantitykind:ElectricFluxDensity ;
+  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
+  rdfs:label "Electric Susceptibility" ;
+.
+unit:SUSCEPTIBILITY_MAG
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  dcterms:description "Magnetic susceptibility is a dimensionless proportionality constant that indicates the degree of magnetization of a material in response to an applied magnetic field. Here M = chi * H. Where M is the magnetization of the material (the magnetic dipole moment per unit volume), measured in amperes per meter, and H is the magnetic field strength, also measured in amperes per meter. Chi is therefore a dimensionless quantity." @en ;
+  qudt:conversionMultiplier 1.0e0 ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (changes requested):
I now realize I need to update our Unit Submission Guide to reflect the fact that we are now including the property qudt:hasDimensionVector in all our Unit instances. Would you mind adding the appropriate values for your new units? (Easily found, of course, via the hasquantitykind/hasDimensionVector path).

Thanks.

nicholascar:
> Would you mind adding the appropriate [qudt:hasDimensionVector] values for your new units?

OK, I've done the best I can here for the moment without perhaps some more examples. 

I used the documentation at [1] of course but it occurs to me that some small updates are needed there (things I couldn't find) but I can't commit to it (since it's a wiki page, not in the repo directly). Can you please provide the general description of the vector before the example, so:

**1. alter the order of listed dimensions to match the vector?**

* Length - meter (m)
* Time - second (s)
* Amount of substance - mole (mole)
* Electric current - ampere (A)
* Temperature - kelvin (K)
* Luminous intensity - candela (cd)
* Mass - kilogram (kg)

 becomes

* Amount of substance - mole (mole)
* Electric current - ampere (A)
* Length - meter (m)
* Luminous intensity - candela (cd)
* Mass - kilogram (kg)
* Temperature - kelvin (K)
* Time - second (s)

**2. Add in this text:**

In QUDT every dimension vector is comprised of each of the above base dimensions and an associated exponent. This way they can be compared easily. An example for the QuantityKindDimensionVector associated with Force is:

becomes

In QUDT every dimension vector is comprised of each of the above base dimensions and an associated exponent in the following way: 
```
(Dimension1 * dimensional_exponent1) * (Dimension2 * dimensional_exponent2) * ...
```
This is repeated for all 7 dimensions and the dimensionless indicator in the following order:

```
Amount of substance, Electric current, Length, Luminous intensity, Mass, Temperature, Time, Dimensionless
```

The dimensions and the dimensionless value are indicated with single letter codes which, for the above dimensions and dimensionless are, in order, as above:

```
A E L I M H T D
```

An example for the QuantityKindDimensionVector associated with Force is:


Finally, I don't think I've nailed the QuantityKind here [3].


[1] https://github.com/qudt/qudt-public-repo/wiki/User-Guide-for-QUDT#3-introducing-dimension-vectors-for-si-units
[2] https://github.com/qudt/qudt-public-repo/wiki/User-Guide-for-QUDT#the-si-dimensions-and-units
[3] https://github.com/qudt/qudt-public-repo/pull/320/files#diff-2149df826d5aade942798543aa01f733022e6040d90c37e60125d55cf5df54c6R8788

steveraysteveray:
@nicholascar, thanks for making it super-easy for me to put your changes into the wiki! That's done.

I will get back to seeing what you did in [3] above as soon as I can.

dr-shorthair:
See #272 for ongoing discussion of relative importance of Unit, Quantity-Kind and DImension-Vector

steveraysteveray (changes requested):
I'm struggling with the idea that unit:SUSCEPTIBILITY_ELEC and unit:SUSCEPTIBILITY_MAG are units rather than QuantityKinds. We already have them defined as quantitykind:ElectricSusceptibility and quantitykind:MagneticSusceptability.

I don't think these two should be added as units.

dr-shorthair:
Elec and mag sus are dimensionless, but observations produce numerical values. Those values have a 'unit of measure' which is conventionally known as the Electrical Susceptibility unit, and the Magnetic Susceptibility unit, respectively. There could be a scale factor (though the conventional one does not). My suggested change would only be to the label.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -18513,6 +18528,30 @@ unit:SurinamDollar
...
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." ;
+  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionOffset "0.0"^^xsd:double ;
+  qudt:latexDefinition "\\chi_{\\text{e}} = \\frac{{\\mathbf P}}{\\varepsilon_0{\\mathbf E}}"^^qudt:LatexString ;
+  qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
+  qudt:hasQuantityKind quantitykind:Dimensionless ;
+  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
+  rdfs:label "Electric Susceptibility" ;
```
dr-shorthair, line 18541: Suggest 
```
  rdfs:label "Electric Susceptibility Unit" ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -18513,6 +18528,30 @@ unit:SurinamDollar
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Surinam Dollar" ;
 .
+unit:SUSCEPTIBILITY_ELEC
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." ;
+  qudt:conversionMultiplier 1.0e0 ;
```
dr-shorthair, line 18535: Suggest using decimal
```
  qudt:conversionMultiplier 1.0 ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -18513,6 +18528,30 @@ unit:SurinamDollar
...
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Magnetic susceptibility is a dimensionless proportionality constant that indicates the degree of magnetization of a material in response to an applied magnetic field. Here M = chi * H. Where M is the magnetization of the material (the magnetic dipole moment per unit volume), measured in amperes per meter, and H is the magnetic field strength, also measured in amperes per meter. Chi is therefore a dimensionless quantity." ;
+  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionOffset "0.0"^^xsd:double ;
+  qudt:latexDefinition "\\chi_\\text{v} = \\frac{\\mathbf{M}}{\\mathbf{H}}"^^qudt:LatexString ; 
+  qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
+  qudt:hasQuantityKind quantitykind:Dimensionless ;
+  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
+  rdfs:label "Magnetic Susceptibility" ;
```
dr-shorthair, line 18553: Suggest 
```
  rdfs:label "Magnetic Susceptibility Unit" ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -18513,6 +18528,30 @@ unit:SurinamDollar
...
+  qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
+  qudt:hasQuantityKind quantitykind:Dimensionless ;
+  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
+  rdfs:label "Electric Susceptibility" ;
+.
+unit:SUSCEPTIBILITY_MAG
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Magnetic susceptibility is a dimensionless proportionality constant that indicates the degree of magnetization of a material in response to an applied magnetic field. Here M = chi * H. Where M is the magnetization of the material (the magnetic dipole moment per unit volume), measured in amperes per meter, and H is the magnetic field strength, also measured in amperes per meter. Chi is therefore a dimensionless quantity." ;
+  qudt:conversionMultiplier 1.0e0 ;
```
dr-shorthair, line 18547: Suggest using decimal
```
  qudt:conversionMultiplier 1.0 ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -18513,6 +18528,30 @@ unit:SurinamDollar
...
+  qudt:conversionOffset "0.0"^^xsd:double ;
```
dr-shorthair, line 18548: Suggest using decimal
```
  qudt:conversionOffset 1.0 ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -18513,6 +18528,30 @@ unit:SurinamDollar
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Surinam Dollar" ;
 .
+unit:SUSCEPTIBILITY_ELEC
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Electric susceptibility is a dimensionless proportionality constant that indicates the degree of polarization of a dielectric material in response to an applied electric field. Here P = epsilon_0 * chi_e * E. Where epsilon_0 is the electric permittivity of free space (electric constant), P is the polarization density of the material chi_e is the electric susceptibility and E is the electric field." ;
+  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionOffset "0.0"^^xsd:double ;
```
dr-shorthair, line 18536: Suggest using decimal
```
  qudt:conversionOffset 1.0 ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -8777,6 +8777,21 @@ unit:KiloGM-M2
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Kilogram Meter Squared" ;
 .
+unit:KiloGM-PER-M3
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Kilogram per cubic metre is an SI derived unit of density, defined by mass in kilograms divided by volume in cubic metres. The official SI symbolic abbreviation is \\(kg \\cdot m^{-3}\\), or equivalently either \\(kg/m^3\\)."^^qudt:LatexString ;
+  qudt:conversionMultiplier 1.0e0 ;
+  qudt:conversionOffset "0.0"^^xsd:double ;
```
dr-shorthair, line 8785: Suggest using decimal
```
  qudt:conversionOffset 1.0 ;
```

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -8777,6 +8777,21 @@ unit:KiloGM-M2
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Kilogram Meter Squared" ;
 .
+unit:KiloGM-PER-M3
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  qudt:plainTextDescription "Kilogram per cubic metre is an SI derived unit of density, defined by mass in kilograms divided by volume in cubic metres. The official SI symbolic abbreviation is \\(kg \\cdot m^{-3}\\), or equivalently either \\(kg/m^3\\)."^^qudt:LatexString ;
+  qudt:conversionMultiplier 1.0e0 ;
```
dr-shorthair, line 8784: Suggest using decimal
```
  qudt:conversionMultiplier 1.0 ;
```

dr-shorthair:
Better to express all multipliers and offsets as `xsd:decimal` - see https://github.com/qudt/qudt-public-repo/issues/307

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Addressed by Comment https://github.com/qudt/qudt-public-repo/pull/320/commits/78ef2c191aee473ba38b69864919bb5c448a17be

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Addressed by Comment https://github.com/qudt/qudt-public-repo/pull/320/commits/78ef2c191aee473ba38b69864919bb5c448a17be

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Addressed by Comment https://github.com/qudt/qudt-public-repo/pull/320/commits/78ef2c191aee473ba38b69864919bb5c448a17be

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Addressed by Comment https://github.com/qudt/qudt-public-repo/pull/320/commits/78ef2c191aee473ba38b69864919bb5c448a17be

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Addressed by Comment https://github.com/qudt/qudt-public-repo/pull/320/commits/78ef2c191aee473ba38b69864919bb5c448a17be

nicholascar:
@dr-shorthair said "Suggest using decimal" for `qudt:conversionMultiplier` x 3:

I don't want to change from the 1.0e0 as implemented as this is more inline with current QUDT examples than 1.0 would be.

There appears to be inconsistency with exponent notation use, e.g. for `unit:BTU_IT-PER-FT3` we have capital E used as such:

`unit:BTU_IT-PER-FT3 qudt:conversionMultiplier 3.725894579E04 ;`

For `unit:BTU_IT-PER-LB_F-DEG_F` we have lower case e used as such:

`unit:BTU_IT-PER-LB_F-DEG_F qudt:conversionMultiplier 4.1868e+03 ;`

I think it's out of scope for this PR to either break convention or change all notation in the vocab, so I leave it to another PR (from someone else, unless you ask nicely).

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Not resolved, see comment below

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Not resolved, see comment below

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
nicholascar: Not resolved, see comment below

dr-shorthair:
There is a mixture of number types in the current catalogue. There is no consistency, so we may as well use best practice for any additions. 

And apart from the valid issues raised by @jmkeil, a TTL file is a text representation after all, so IMHO has no business telling a processor to use some (inconsistent) number of bits for the representation!

nicholascar:
> There is a mixture of number types in the current catalogue...

OK, I have changes the 4 unit additions I've made to use decimals for their `qudt:conversionMultiplier` properties.

dr-shorthair:
(To be quite honest, I can't see any compelling use-cases for `xsd:double` and `xsd:float`. Double and float mean different things in different systems, and XML (and TTL/N3 and JSON) are text representations after all.)

nicholascar:
Not sure I follow: by using the un-quoted form of `1.0` & `0.0` that I have, I've encoded in the Turtle that these are decimals, so the equivalent to using the explicit `"1.0"^^xsd:decimal` etc. Since Turtle data is declarative, each RDF store etc. will (should) follow this instruction to encode the value in the way indicated - with a decimal datatype (or its underlying equivalent), so the type matters and they are not just "text representations after all".

jhodgesatmb:
Wasn’t this bug report about the type of conversion multipliers? How did susceptibility come into it? I don’t recognize it at all, and wonder why it is all caps. Anyway, let’s talk about it.

Jack

nicholascar:
> why it is all caps

I was following what I thought was precedence, e.g. `unit:OERSTED` & `unit:PERMITTIVITY_REL` but I'm not attached to caps at all so would you prefer `MagneticSusceptibility` & `ElectricSusceptibility`?

jhodgesatmb:
No, you are right, that (all caps) is the qname convention. Sorry.

Jack Hodges, Ph.D.
Arbor Studios

steveraysteveray:
OK, so to summarize where we are at this point, here are my observations (and my positions):
1. I'm convinced by @nicholascar in stating conversionMultiplier and conversionOffset values as unquoted decimals, as a valid alternative to the explicitly quoted xsd:decimal declaration.
2. I'm willing to accept the two units for magnetic and electric susceptibility, with the caveat that they point to the respective magnetic and electric susceptibility quantitykinds rather than dimensionless.
3. I agree we need a separate Issue to consolidate up our various representations of numbers toward xsd:decimal or unquoted decimals.

Sound OK?

jhodgesatmb:
Yes

dr-shorthair:
> I agree we need a separate Issue to consolidate up our various representations of numbers toward xsd:decimal or unquoted decimals.

It is already there - #307 

Note that ['unquoted decimal' is merely syntactic-sugar for `"NN.N"^^xsd:decimal`](https://www.w3.org/TR/turtle/#abbrev) - i.e. to an RDF parser there is no difference. 

To clarify my [grouchy comments](https://github.com/qudt/qudt-public-repo/pull/320#issuecomment-749290196) about float and double: The main point is that  XML/JSON/TTL/N3 is a textual transfer format, so `xsd:decimal` is 'honest' and also it allows for arbitrary precision. The comments by @jmkiel on `0.1` are also very relevant - if N*0.000001*1000000 does not reliably = N then we have a problem! 

In general, telling the processor what internal representation to use after parsing is intrusive and unpredictable, since it can produce different results in different architectures (though I note that `xsd:double` is defined as IEEE 64-bit, so maybe that objection is stale).

dr-shorthair:
Fix now submitted as PR https://github.com/qudt/qudt-public-repo/pull/311

nicholascar:
> In general, telling the processor what internal representation to use after parsing is intrusive and unpredictable

Disagree: you - the data definer - _should_ set the expectation for the literal. The "internal representation" isn't `double` `float` etc., that is an external representation. The _internal_ representation is however the system implements float, double etc. For example, in Python, there is no difference between a `float` or a `double` [1] so a Python RDF system, e.g. RDFlib, would retain `xsd:float` or `xsd:double` in the RDF literal type it stores ("external") but the "internal" representaion - a a 64 signed byte - would be the same for each.


[1] https://stackoverflow.com/questions/33960828/python-float-or-double

nicholascar:
I support having everything in decimal though!

...and it looks like this PR is ready to go?

steveraysteveray (approved):

PR #321, opened 2020-12-17
Loop3d community

nicholascar:
A community profile of QUDT by and for the Loop3D community.

This is a simple profile containing a single SKOS-only vocabulary of 7 units. There are two files because one is the vocab and the other is the profile definition. The vocab is therefore characterised as a _Profiles Vocabulary_ profile `Resource`.

This is a simple subset list profile but the vocab will grow with the intention being that the Loop3D project can keep updating their profile and use QUDT elements directly (as SKOS Concepts but retaining the original URIs) rather than reinventing Units.

**NOTE**: that several of the units here are not yet in QUDT but are described within a Pull Request #320 this this PR is dependent on that one being merged first to be legitimate.


steveraysteveray (commented):
While you caught the "a a demominator", (thanks for that), let's go ahead and make it "denominator"

nicholascar:
> While you caught the "a a demominator", (thanks for that), let's go ahead and make it "denominator"

Done! I may have only been able to change them within a small part of the ontology in this PR as I've edited the files online for this update, so you might have to check again for "demominator" after this still.

steveraysteveray:
I'm assuming that we're waiting on this one until #320 is completed. At that point, I'm hoping @nicholascar will resolve the merge conflict.

nicholascar:
Conflicts fixed, awaiting PR #320 (which I think is about ready!)

steveraysteveray (approved):
Not a show-stopper, but you changed the text in a few dcterms:description fields from "...square <unit>..." to "...<unit squared...". 
This is contrary to the convention we adopted for rdfs:label [here](https://github.com/qudt/qudt-public-repo/wiki/Unit-Vocabulary-Submission-Guidelines#conventions-for-rdfslabel).
The reason for this convention is because it scales more naturally for higher orders (e.g. cubic, quartic, quintic...) while there aren't good words beyond squared and cubed.
Since we don't insist on the same convention for dcterms:description, I can live with it, although it is inconsistent.

steveraysteveray:
Sorry, @nicholascar, would you mind resolving the conflicts that just appeared because I merged the decimal representation PR?

nicholascar:
Conflict resolved

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -22510,13 +22510,17 @@ unit:MegaYR
     a qudt:DerivedUnit ;
     qudt:plainTextDescription "1,000,000-fold of the derived unit year." ;
     qudt:conversionMultiplier 1000000.0 ;
```
dr-shorthair, line 22512: Is this the correct multiplier? It should be relative to the SI unit for this dimensionality - i.e. seconds.

nicholascar: Ooops, assumed it was relative to the YR but obviously how could it be? Have changed it to 1M x YR = 31557600000000.0

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -22509,14 +22509,18 @@ unit:MegaYR
...
     qudt:plainTextDescription "1,000,000-fold of the derived unit year." ;
-    qudt:conversionMultiplier 1000000.0 ;
-    qudt:conversionOffset 0.0 ;
+    qudt:conversionMultiplier 31557600000000.0 ;
     qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T1D0 ;
     qudt:hasQuantityKind quantitykind:Time ;
+    qudt:informativeReference "https://en.wiktionary.org/wiki/megayear"^^xsd:anyURI ;
+    qudt:isScalingOf unit:YR ;
+    qudt:prefix <http://qudt.org/2.1/vocab/prefix/Mega> ;
     qudt:symbol "Myr" ;
```
dr-shorthair, line 22518: Also add 
```
  qudt:ucumCode "Ma"^^qudt:UCUMcs ;
```

nicholascar: Done

steveraysteveray (approved):

PR #343, opened 2021-01-23
Srr schema cleanups

steveraysteveray:
Did a bunch of type fixes, typos, etc.
Still to do:
Create a bunch of dimension vectors that are referred to but don't exist (laborious, painful)
Still need to test the effect of OWL2 RL reasoning on our constraint violations


steveraysteveray:
Superseded by the next PR.

PR #349, opened 2021-02-03
 Generated missing dimension vectors referred to by units

steveraysteveray:


dr-shorthair:
These should really be generated on-demand ... ;-)

steveraysteveray:
That's an interesting observation. In the end, I don't think it's practical for the three user communities:
1. The linked data folks. Yes, we could equip our EDG server with the ability to dereference an arbitrary dimension vector name and return a dynamically created instance.
2.  The EDG/SHACL community. Yes, we could create some SHACL shape using sh:value to infer all the properties of a dimension vector.
3. The "industrial use" OWL community. Not so easy for people who want to download the vocabularies and keep them under careful version control in-house. It seems to me that they would like explicit, declared instances of the various dimension vector instances, without resorting to SPIN inferencing (i.e. magic properties).

I have written a SPARQL query to examine any graph for references to dimension vectors that have not been instantiated, and to then update the dimension vector graph with those. That's what I used for this PR. Trivial to run any time, so for now I plan to just use that if needed.

If you have an elegant suggestion that addresses the three communities above, it would be interesting to hear.

dr-shorthair:
I agree that it doesn't work for user community 3. 
Thanks for this analysis.

jhodgesatmb (approved):

PR #352, opened 2021-02-11
Relaxed valueUnion range constraint in OWL, replaced in SHACL

steveraysteveray:


jhodgesatmb (approved):
I am unsure why we don't use the numericUnion datatype that was in the ontology before and perhaps still is, since it seems the same as what you are defining in this union, but if numericUnion is producing problems (and I seem to recall that it did as well) then maybe this is the right thing to do.

PR #374, opened 2021-03-31
Add MicroMOL-PER-M2

### LINKED ISSUE #373: New unit request - MicroMOL-PER-M2
Required for some oceanography data

dr-shorthair:
Fixes #373


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -14068,6 +14085,7 @@ unit:MicroMOL-PER-MicroMOL-DAY
   qudt:conversionMultiplier 0.0000115740740740741 ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T-1D0 ;
```
steveraysteveray, line 14086: This is an interesting case. The way you did it is fine with me. An alternative is to add a qkdvNumerator of A1 and a qkdvDenominator of A1...T-1. I'm not sure if it adds any value though.

We don't have any rules about how to split up mixed ratios like this.

steveraysteveray (approved):

PR #385, opened 2021-04-15
Added missing isDefinedBy links

### LINKED ISSUE #384: PSU not in the units catalogue
http://qudt.org/vocab/unit/PSU does not appear in http://www.qudt.org/doc/DOC_VOCAB-UNITS.html 
Is that because it is 'not recommended' ?

dr-shorthair:
Fixes #384


steveraysteveray (approved):
Thanks, Simon!

PR #391, opened 2021-04-29
fix wrong values for qudt:uneceCommonCode

Toni100:
discovered by Wikidata cross-checks

I verified the corrections by looking at the latest revision 16 of recommendation 20 on the official website:
https://unece.org/trade/uncefact/cl-recommendations

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -1082,7 +1082,7 @@ unit:BIT
   qudt:omUnit <http://www.ontology-of-units-of-measure.org/resource/om-2/bit> ;
   qudt:symbol "bit" ;
   qudt:ucumCode "bit"^^qudt:UCUMcs ;
-  qudt:uneceCommonCode "J63" ;
+  qudt:uneceCommonCode "A99" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Bit"@en ;
 .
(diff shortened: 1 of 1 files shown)


jhodgesatmb:
It is easier to check if you say which units your source says are incorrect and why. Otherwise we have to check everything.

Toni100:
The link I provided is the authoritative source who produces those identifiers.
The units which have been changed can be seen in the diff. Would that be enough?
I used Wikidata to see which values might be wrong (Wikidata is not a primary source), and then fixed the ones which were actually wrong by looking into the official document.

jhodgesatmb:
I ask partly because the UNECE codes were provided to us or acquired by us directly from the source, so have they changed their codes? Secondly, my request was not about the source of information but of the comment in the PR. If the PR comment tells me to look at a specific list of units then the review is easier.

steveraysteveray:
Thanks, Toni! I reviewed all 5 of your suggested changes, and you are absolutely correct. The only exception is that I couldn't find A99 (bit) in the spreadsheet. Was that in some secret place?

Toni100:
@jhodgesatmb: Take for instance `D27`, which was present on both `unit:M2-PER-SR` and `unit:SR` - two dimensionally incompatible units. This looks pretty clearly like a typo. (That might have been wrong in the source, too, but confirming that would require some archeology on the older source revisions - maybe not worth our time.)

@steveraysteveray: `A99` appears in the third table (Annex II & Annex III). For some weird reason there is a filter activate on this table. Disabling it (menu Data -> AutoFilter in LibreOffice) makes all entries appear.

Toni100:
PS: Looking for codes which appear on different units which are not `owl:sameAs` might be a way to discover more such typos.

steveraysteveray:
Got it! Thanks.

steveraysteveray (approved):

Issue #394, opened 2021-05-14
Units with ?/?? typo in rdfs:label


mkarpisek:
In qudt 2.1.10 (also current master) there is several units with ?/?? in their rdfs:label (both en/en-us), which looks like an error.

Seems like they should be rather "Kilogram Force.." and "Vietnamese Dong"?

#### Table of affected units
| unit                                              | rdfs:label                                 |
|---------------------------------------------------|--------------------------------------------|
| http://qudt.org/vocab/unit/KiloGM_F-M             | Kilogram?force Metre                       |
| http://qudt.org/vocab/unit/KiloGM_F-M-PER-SEC     | Kilogram?force Metre Per Second            |
| http://qudt.org/vocab/unit/KiloGM_F-M-PER-CentiM2 | Kilogram?force Metre Per Square Centimetre |
| http://qudt.org/vocab/unit/VietnameseDong         | Vietnamese ??ng                            |

#### Sparql used to get affected units
```
PREFIX qudt: <http://qudt.org/schema/qudt/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

SELECT *
WHERE {
    ?uri rdf:type qudt:Unit .
    ?uri rdfs:label ?label.  
    FILTER (CONTAINS(?label, "?")).
}
ORDER BY ?label
```

jhodgesatmb:
Those should be underscores. Some scripting translation error I suspect.  Thanks for pointing this out.

PR #395, opened 2021-05-14
Fixes ?/?? typo in units rdf:label

mkarpisek:
corrected labels are aligned with other existing similar unit desciptions
* Kilogram?force.. => Kilogram Force..
* Vietnamese ??ng => Vietnamese Dong

Fixes qudt/qudt-public-repo#394


jhodgesatmb:
Our qname convention is to use underscores to qualify a unit. The label should have reflected that.

steveraysteveray (approved):
The amended labels look good. Jack is correct that the URIs contain the underscore, but the labels don't.

Thanks.

PR #398, opened 2021-05-23
Updates to UNECE codes and update of KiloGM-PER-KiloMOL

stuchalk:
I have added some additional UNECE codes, corrected some and updated the unit KiloGM-PER-KiloMOL for its ucum code, description, and label.


steveraysteveray:
Thanks, Stuart. It's with a little trepidation that I'm approving the 140 changes you made to the dcterm descriptions, removing the extra blank. There's no way you would know that we're in the midst of a scrub of all our dcterm descriptions for latex errors. I'm hoping that we don't get a large number of merge conflicts when we finish our scrub. 

Regardless, your fix of that makes sense.
I also spot checked your UNECE codes against rec20_Rev15e-2020.xls. Is that the UNECE publication you used? Seems correct. Do you have any knowledge of what the UNECE folks mean by an 'X' in the Status column of Annex II/III? Specifically, you added code D70 for the IT calorie, which also shows code R4. I'm confused as to which code should be used.

steveraysteveray (approved):

PR #414, opened 2021-06-11
added unit KiloGM-PER-M-SEC. 

kitchenprinzessin3880:
This unit can be used to express pressure, besides N/m2.


steveraysteveray (approved):

steveraysteveray:
Wow, your typo in the commit message threw me off, because KiloGM-PER-M-SEC does not have the dimensionality of pressure. However, you are correct that KiloGM-PER-M-SEC2 does have the correct dimensionality, which is your change in the actual file. And we already have KiloGM_F-PER-M2 as a unit of pressure, which is quite different from KiloGM-PER-M2 which is a unit of surface density.

I'm rambling a bit because a Kilogram of mass and a Kilogram of force might be confused by some to be the same thing, and they most definitely are not.

kitchenprinzessin3880:
my commit message from the source tree somehow got truncated, so the last char '2' was missing from the title. apologies for the inconvenience caused.

PR #434, opened 2021-09-20
commit for M4M PR2

stuchalk:
Added units of mass concentration NanoGM-PER-MilliL and MicroGM-PER-MilliL.  This is for the upcoming Metadata for Machines (M4M) workshop to be run by GO-FAIR.  The M4M workshop is focused on immunological lab analyses where this unit is prevalent.  These are additional common units as reported in the NCATS "[Assay Guidance Manual](https://www.ncbi.nlm.nih.gov/books/NBK92434/)' in section 'Immunoassay Parameters' (version date July 1, 2021).


steveraysteveray (approved):
This seems fine, although it resurfaces the question we face about aliases. Currently, MassConcentration and MassDensity mean the same thing, along with the common interpretation of Density. The question remains whether we should have all three as separate, cross-referencing instances, and if so, is owl:sameAs the best predicate? If not, should we be using skos:altLabel for two of the names?

PR #442, opened 2021-10-29
 Expanded the definition of qudt:inferredApplicableUnit to include ow…

steveraysteveray:
…l:sameAs quantitykinds


steveraysteveray:
Problem is more complicated that originally appeared.

Issue #454, opened 2022-01-06
best practice for using versioned iris in relation from unit to prefix


fkleedorfer:
Some Units have prefixes, e.g.

```
<http://qudt.org/vocab/unit/KiloA> <http://qudt.org/schema/qudt/prefix> <http://qudt.org/2.1/vocab/prefix/Kilo> .
```

However, their IRIs are 'versioned', i.e., they contain the ontology version number. These IRIs are nowhere to be found in the prefixes file accessible at https://qudt.org/2.1/vocab/prefix. 

In contrast, the relationship between `Unit` and `QuantityKind` is not 'versioned' in this way. 

For my use case, I download the ttl files I need (unit, quantitykind, prefix) and put the data into a local rdf store. When I try to join `Unit` and `Prefix` the attempt fails (obviously, once you know). My quick fix is to change the `qudt:prefix` uris that the unit entitys point to, but that's a hassle as I have to do this with every update. What's the best practice here?

steveraysteveray:
Florian,
Thanks so much for catching this! We hadn't really explicitly used the prefix individuals much, so the reference from the units slipped by unnoticed. It is definitely an error - luckily easily fixed with a global replace to remove the version number. Please stay tuned for an update.

Issue #455, opened 2022-01-06
qudt:isScalingOf not applied consistently


fkleedorfer:
I had high hopes to find the base unit for any scaled unit, but it seems it's not in the data.

E.g.
```
unit:DeciM qudt:isScalingOf unit:M ;
unit:CentiM qudt:isScalingOf unitM;
```
but
the triple is missing for 
```
unit:KiloM 
unit:MilliM 
```
Many other examples can be found. 

Having complete scaling and prefix information would be very useful, it also seems that the triples can be generated automatically based on the unit's localname.

fkleedorfer:
Here's a CONSTRUCT query that generates all the missing scaling/prefix triples (I believe), for base units.
Note:
* Information measures (`unit:Byte`) etc are not included because the base units are selected by their multiplier of 1, but byte has a different one.
* Weights are arguably misleading as the base SI unit is `unit:KiloGM`, so all prefixes of scaled units have a wrong multiplier as using them would convert toward grams, not kilograms

Anyway: 
```
PREFIX qudt: <http://qudt.org/schema/qudt/>
PREFIX unit: <http://qudt.org/vocab/unit/>
PREFIX prefix: <http://qudt.org/vocab/prefix/>
PREFIX kind: <http://qudt.org/vocab/quantitykind/>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

CONSTRUCT {
    ?scaledUnit
        qudt:isScalingOf ?baseUnit ;
        qudt:prefix ?prefix .
}
WHERE
{
    {
        # select all base units (multiplier 1, not using prefixes, except for KiloGM)
        ?baseUnit a qudt:Unit ;
            qudt:conversionMultiplier 1.0 .
            BIND(REPLACE(STR(?baseUnit), "^.+/", "") as ?baseUnitLocalName)
            FILTER(! CONTAINS(?baseUnitLocalName, "-"))
    }

    ?scaledUnit a qudt:Unit ;
    BIND(REPLACE(STR(?scaledUnit), "^.+/", "") as ?scaledUnitLocalName)
    ?prefix2
        a qudt:Prefix ;
        rdfs:label ?prefixLabel2 .
    FILTER EXISTS {

         FILTER (
            ?scaledUnitLocalName = CONCAT(STR(?prefixLabel2), ?baseUnitLocalName)
[...]

PR #456, opened 2022-01-06
Implementing both suggestions made by @fkleedorfer

steveraysteveray:


steveraysteveray:
This resolves #454 and #455.

jhodgesatmb (approved):

fkleedorfer:
That was fast ;-)

I found out that with that query I provided, KiloGM does not get its prefix linked. So if you are in the mood, that one triple could be added:
```
unit:KiloGM qudt:prefix prefix:Kilo .
```
(hopefully no typos here)

PR #460, opened 2022-01-16
added unit cm3/cm3

kitchenprinzessin3880:
excluded the properties uneceCommonCode and iec61360Code as there is no relevant value found for the unit added; see
https://unece.org/fileadmin/DAM/cefact/recommendations/bkup_htm/add3a.htm
https://cdd.iec.ch/cdd/iec61360/iec61360.nsf/Units/0112-2---62720%23UAA394?opendocument


steveraysteveray:
Wouldn't the conversion multiplier just be 1.0? This unit is equal to M3-PER-M3, yes?

kitchenprinzessin3880:
@steveraysteveray - copy paste mistake (huhu) thanks for spotting it - i updated the dimension ;)

steveraysteveray (approved):

PR #468, opened 2022-02-09
 Added ventilation rate per floor area

steveraysteveray:


jhodgesatmb (commented):
unit:L-PER-SEC-M2 should have an M-2 in its dimension vector shouldn't it?

I am not sure about the L/T (some kind of distance rate) for the VentilationRatePerFloorArea. It seems to me that ventilation would be more of a volume rate than a distance rate, but maybe volume over area is length and that is how you did it.

steveraysteveray:
You are correct. Volume/Area gave me Length.
The only alternative is to use numerator and denominators to make it more explicit, but I think that is a slippery slope except for dimensionless units.

dr-shorthair:
Length is a typical unit for fluxes. E.g. rainfall, or irrigation, which conceptually might be easier thought of as volume per area. 
Conversely, you can see some pretty funky units for volume coming back at you - e.g. 'acre-feet'. 
What fun!

jhodgesatmb (approved):

PR #469, opened 2022-02-10
Units and quantity kinds added by https://agrimetrics.co.uk/

nimonika:
Units and quantity kinds added by https://agrimetrics.co.uk/. These units are widely used across the agri-food sector.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -23627,47 +23627,6 @@ unit:M2-PER-M2
     <http://qudt.org/schema/qudt/symbol> "m^2/m^2" ;
     rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> .
 
-unit:SHANNON_DIVERSITY
-    a qudt:Unit, qudt:LogarithmicUnit ;
-    rdfs:label "Shannon Diversity Index"@en ;
-    qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
-    qudt:plainTextDescription "The uncertainty of predicting the species of an individual sampled from a given sampling region. A measure of the entropy of the species sampled in the region. The index may be taken using 2, e or 10 as the base of the logarithm; these three measures must be scaled appropriately in order to be comparable." ;
-    qudt:hasQuantityKind quantityKind:ShannonDiversityIndex ;
-    rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> .
-
-unit:SHANNON_DIVERSITY_LOG10
-    a qudt:Unit, qudt:LogarithmicUnit ;
-    qudt:conversionMultiplier "3.3219280949"^^xsd:double ;
-    qudt:conversionOffset "0"^^xsd:double ;
-    rdfs:label "Shannon Diversity Index (base 10)"@en ;
-    qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
-    qudt:plainTextDescription "The uncertainty of predicting the species of an individual sampled from a given sampling region. A measure of the entropy of the species sampled in the region." ;
-    qudt:hasQuantityKind quantityKind:ShannonDiversityIndex ;
-    rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> .
-    skos:broader unit:SHANNON_DIVERSITY .
-
-unit:SHANNON_DIVERSITY_LOGE
-    a qudt:Unit ;
-    qudt:conversionMultiplier "1.4426950409"^^xsd:double ;
-    qudt:conversionOffset "0"^^xsd:double ;
-    rdfs:label "Shannon Diversity Index (base e)"@en ;
-    qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
-    qudt:plainTextDescription "The uncertainty of predicting the species of an individual sampled from a given sampling region. A measure of the entropy of the species sampled in the region." ;
-    qudt:hasQuantityKind quantityKind:ShannonDiversityIndex ;
-    rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
-    skos:broader unit:SHANNON_DIVERSITY .
-
-unit:SHANNON_DIVERSITY_LOG2
-    a qudt:Unit ;
-    qudt:conversionMultiplier "1"^^xsd:double ;
-    qudt:conversionOffset "0"^^xsd:double ;
-    rdfs:label "Shannon Diversity Index (base 2)"@en ;
-    qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
-    qudt:plainTextDescription "The uncertainty of predicting the species of an individual sampled from a given sampling region. A measure of the entropy of the species sampled in the region." ;
-    qudt:hasQuantityKind quantityKind:ShannonDiversityIndex ;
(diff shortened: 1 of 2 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -23602,3 +23602,106 @@ THE UCUM TABLE (IN ALL FORMATS), UCUM DEFINITIONS, AND SPECIFICATION ARE PROVIDE
...
+    qudt:conversionOffset "0.0"^^xsd:double ;
+    qudt:hasDimensionVector qkdv:A0E0L0I0M0H1T1D0 ;
+    qudt:plainTextDescription "The sum of excess temperature over 5.5°C, where the temperature is the mean of the minimum and maximum atmospheric temperature in a day. This measure is appropriate for most cereal crops." ;
+    qudt:hasQuantityKind quantityKind:GrowingDegreeDay_Cereal ;
+    rdfs:isDefinedBy <https://data.agrimetrics.co.uk/ontologies/qudt-extension> .
+
+unit:M2-PER-M2
+    a  qudt:Unit;
+    rdfs:label "square metre per square metre" ;
+    qudt:hasDimensionVector qkdv:A0E0L2I0M0H0T0D0 ;
```
steveraysteveray, line 23619: This should be dimensionless, i.e. qkdv:A0E0L0I0M0H0T0D1. To distinguish this dimensionless ratio from others, I recommend you also include:
  qudt:qkdvDenominator qkdv:A0E0L2I0M0H0T0D0 ;
  qudt:qkdvNumerator qkdv:A0E0L2I0M0H0T0D0 ;

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -23602,3 +23602,106 @@ THE UCUM TABLE (IN ALL FORMATS), UCUM DEFINITIONS, AND SPECIFICATION ARE PROVIDE
...
+    qudt:hasQuantityKind quantityKind:AreaRatio ;
+    qudt:conversionMultiplier "1"^^xsd:double ;
+    qudt:conversionOffset "0.0"^^xsd:double ;
+    qudt:plainTextDescription "A square metre unit of area per square metre"^^rdf:HTML ;
+    <http://qudt.org/schema/qudt/symbol> "m^2/m^2" ;
+    rdfs:isDefinedBy <https://data.agrimetrics.co.uk/ontologies/qudt-extension> .
+
+unit:SHANNON_DIVERSITY
```
steveraysteveray, line 23627: If you never want users to instantiate this (abstract) unit, you can remove the qudt:hasQuantityKind triple. Then only the specific ones (LOG10, etc.) would show up as applicableUnit when examining the ShannonDiversityIndex.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -23602,3 +23602,106 @@ THE UCUM TABLE (IN ALL FORMATS), UCUM DEFINITIONS, AND SPECIFICATION ARE PROVIDE
...
+    qudt:conversionMultiplier "1"^^xsd:double ;
+    qudt:conversionOffset "0"^^xsd:double ;
+    rdfs:label "Shannon Diversity Index (base 2)" ;
+    qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
+    qudt:plainTextDescription "The uncertainty of predicting the species of an individual sampled from a given sampling region. A measure of the entropy of the species sampled in the region." ;
+    qudt:hasQuantityKind quantityKind:ShannonDiversityIndex ;
+    rdfs:isDefinedBy <https://data.agrimetrics.co.uk/ontologies/qudt-extension> ;
+    skos:broader unit:SHANNON_DIVERSITY .
+
+unit:FRACTION
```
steveraysteveray, line 23668: Interesting. Nice addition!

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -23602,3 +23602,106 @@ THE UCUM TABLE (IN ALL FORMATS), UCUM DEFINITIONS, AND SPECIFICATION ARE PROVIDE
...
+    qudt:conversionOffset "0"^^xsd:double ;
+    qudt:hasDimensionVector qkdv:A0E0L0I0M1H0T-2D0 ;
+    qudt:plainTextDescription "A unit of energy per unit area, equivalent to 3 600 000 joules per square metre." ;
+    qudt:hasQuantityKind quantityKind:EnergyPerArea ;
+    rdfs:isDefinedBy <https://data.agrimetrics.co.uk/ontologies/qudt-extension> .
+
+unit:IU-PER-MilliGM
+    a qudt:Unit ;
+    rdfs:label "International Unit per milligram" ;
+    qudt:conversionMultiplier "0.001"^^xsd:double ;
```
steveraysteveray, line 23701: Shouldn't the multiplier be 0.000001?

steveraysteveray (changes requested):
A few errors slipped in. See embedded comments.

Also, could I ask that all your rdfs:label values add @en after the string for the units?
If you really want to be nice, you could also include the U.S. spellings (meter instead of metre) as a second triple, with the @en-us tag.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
steveraysteveray: (The applicableUnit generation algorithm is documented [here](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#4-computing-applicable-units-for-a-quantitykind).)

darkling: Hi Steve, 

I'm Monika's colleague.

The uses of this unit that I've seen usually don't specify which base they've used for the logarithm, but it does appear that multiple bases are in use -- I wasn't able to pin down any specific recommendations about which one should be used. We left the "abstract" version in so that it's possible to indicate the type of unit without having to specify a base that's undocumented by the upstream generators of the data we're trying to annotate.

If that deliberate vagueness is against the intent of QUDT, then by all means we can drop the qudt:hasQuantityKind, but it's useful to have it, I think.

steveraysteveray: Here's what I propose, which is what we did for unit:OZ and others. The ounce is a measure of force, mass, and volume. We chose what we believed to be the "most common" one - mass - and used unit:OZ to mean that. The others are named unit:OZ_F and unit:OZ_VOL (with yet a further qualifier of _UK or _US). The point was that the unqualified version means **_something_** unambiguous, which can be interrogated.

In your case, we would collectively choose a "most common" logarithmic scale, and anoint unit:SHANNON_DIVERSITY to mean that. The other variants would have the qualifier. The description would make each one clear. You may also want to include:
rdfs:seeAlso <the other variants> 
... in each one, but that is your choice.

nimonika: @steveraysteveray Thanks. We have decided that for now, we will not submit the Shannon index units. We will make another submission on this, once we get some more clarity at our end.

steveraysteveray: OK. Another option you might want to consider in the meantime is to add the three qualified versions which would support applications where the scale is known.

steveraysteveray (approved):

PR #471, opened 2022-02-10
Add reciprocal nanometer, kilometer, micrometer, and millimeter

jat255:
Apologies if this is not exactly right, as I've just copied from the `unit:PER-M` definition (and I've never edited an ontology definition before).

Happy to make any edits needed


steveraysteveray (changes requested):
Nice job! Just one tiny request - there is a repeated "a a" in each description. Not a show-stopper, but it would be nice to catch it before the merge.

jat255:
Good spot. That "a a denominator" came from the "PER-M" definition that I copy/pasted. I fixed it with a find replace and it caught a few other instances of the error.

PR #474, opened 2022-02-17
Fix quantitykind prefix case for agrimetrics units

bkr-iotic:
restores parsability to affected ttl's


steveraysteveray (approved):

steveraysteveray:
Thanks for catching this!

PR #484, opened 2022-03-30
 Allow decimal, float, double for conversion factors

steveraysteveray:


jhodgesatmb (approved):

mgberg:
This effectively reverts #307/#311. Could someone provide some insight into why this decision was made? Additionally, has there been any thought in enabling conversion factors provided as a fraction (for example, instead of `qudt:conversionMultiplier` use something like `qudt:conversionNumerator` and an optional `qudt:conversionDenominator` assumed to be `1` if not present) to represent ratio conversion factors instead of a decimal to arbitrary precision?

jhodgesatmb:
There has been discussion over the years to change conversion multipliers to handle more general relations but it hasn’t happened yet. It is not a simple change.

Jack Hodges, Ph.D.
Arbor Studios

mgberg:
Good to know. What about https://github.com/qudt/qudt-public-repo/issues/307 / https://github.com/qudt/qudt-public-repo/pull/311 - Can you provide some insight into why that decision was reverted?

PR #493, opened 2022-05-03
Added two new units: umol/g and umol/gh

oibeal:
Two new units added: 
  1. umol/g
  2. umol/g/h


steveraysteveray (approved):

dr-shorthair:
Making these convertable with the commonly-used alternative mass-based units requires a source of molecular-mass data. DOes QUDT have a story about how to do this?

steveraysteveray:
Of course, we cannot be responsible for guiding people to all related standard quantities, but I would go to an authoritative source such as [this site](https://webbook.nist.gov/chemistry/mw-ser/) at NIST.

PR #511, opened 2022-06-23
request to add new unit unit:MilliGM-PER-Hectare

elisagirola:


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -15363,6 +15363,17 @@ unit:MilliGM-PER-CentiM2
   rdfs:label "Milligram Per Square Centimeter"@en-us ;
   rdfs:label "Milligram Per Square Centimetre"@en ;
 .
+unit:MilliGM-PER-Hectare
+  a qudt:Unit ;
+  qudt:conversionMultiplier 0.00000000001 ;
```
steveraysteveray, line 15368: I believe this should be 10^-10, not 10^-11. So it should be 0.0000000001

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -15363,6 +15363,17 @@ unit:MilliGM-PER-CentiM2
   rdfs:label "Milligram Per Square Centimeter"@en-us ;
   rdfs:label "Milligram Per Square Centimetre"@en ;
 .
+unit:MilliGM-PER-Hectare
```
steveraysteveray, line 15366: This should be unit:MilliGM-PER-HA

steveraysteveray (changes requested):

steveraysteveray (approved):

PR #517, opened 2022-07-06
add a label for TONNE-PER-HA-YR

jbkoh:
`unit:TONNE-PER-HA-YR` is missing its `rdf:label`. Adding "tonne per hectare year" referring to other examples


steveraysteveray:
Jason,
I'd like to just approve your suggestion, but we're in the middle of #512 created by Lachlan-Charles for this unit. Let's give them a chance to respond to my suggestions on that one without stepping on toes.   :)

PR #525, opened 2022-07-19
Automated deviation check to other ontologies using ABECTO

jmkeil:
Hi. This adds a configuration to automatically compare your ontology with other unit ontologies using [ABECTO](https://github.com/fusion-jena/abecto) after each commit and to return a table of found deviations. Maybe you are interested to use this.


jmkeil:
I wasn't sure where to put the comparison configuration. For some reason it doesn't work if I put it into the workflow folder as I did in https://github.com/HajoRijgersberg/OM/pull/69 (analogous PR for OM).

jmkeil:
On https://github.com/jmkeil/qudt-public-repo/actions/runs/2698908908 you can find a preview how the results will look like. The job is marked as failed, because deviations have been found. Deviations can be handled by
* fixing the own ontology,
* fixing the other ontology,
* fixing the mapping (e.g. https://github.com/qudt/qudt-public-repo/pull/525/commits/6a06b8a402c6a3758314f77d869367cdd395a1fa#diff-e48e569c0c233b06e9a61b27cf0f6ddedcd8142da4c91bf3d2f8e0debbbbc6b9R715), or
* annotating the other ontologies value wrong in the comparison configuration (e.g. https://github.com/qudt/qudt-public-repo/pull/525/commits/6a06b8a402c6a3758314f77d869367cdd395a1fa#diff-e48e569c0c233b06e9a61b27cf0f6ddedcd8142da4c91bf3d2f8e0debbbbc6b9R833, an appropriate snippet is provided in the last column of the result table).

Alternatively, the parameter `--failOnDeviation` in https://github.com/qudt/qudt-public-repo/pull/525/commits/c904f9bc2c7ec852fe9d9affce69593c3e43a8cb#diff-311f6477bc95af7ccedbe3225a241452ddb76b61cea146f6103c84a5059d6fb8R12 can be removed. Then the job will only fail, if the comparison could not be executed for any reason.

jhodgesatmb:
The problem as I see it is that there is no baseline ontology that is accepted as the authoritative standard. As a result, and as you say, the comparison will always fail unless the [2] ontologies merge or otherwise mitigate their differences. For this reason it is difficult for me to see adding this feature to the code base.

Jack Hodges, Ph.D.
Arbor Studios

dr-shorthair:
I'm more optimistic. There are different emphases coming from the different projects, so the ontologies choose different pathways through the problem. Nevertheless the results should be equivalent. 

The tool merely highlights inconsistencies, which provides an _opportunity_ for further investigation. 
Sometimes it will find an error that you had missed, which you can then correct. 

It is informative, and I don't see the harm in running it automatically.

jmkeil:
I had a look into the documentation and it seems possible to [configure GitHub actions for only manually execution](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). In my opinion, automated execution would be better, as manual executions are more likely to not happen at all. At least in software testing, this is best practice. But, of course, this is your project and your decision. What are your preferences?

> We will place the comparison [configuration ~~results~~] into a folder that isn't included in the release.

_Originally posted by @jhodgesatmb in https://github.com/qudt/qudt-public-repo/issues/526#issuecomment-1190637288_

What would be the preferred path of the comparison configuration (not results - these do not get committed into the repository)?

> The problem as I see it is that there is no baseline ontology that is accepted as the authoritative standard.

@jhodgesatmb: We had this discussion earlier. [My answer](https://github.com/qudt/qudt-public-repo/issues/307#issuecomment-740502498) is still the same. (And I don't have a problem with sticking to our different viewpoints.)

> The tool merely highlights inconsistencies, which provides an opportunity for further investigation.
Sometimes it will find an error that you had missed, which you can then correct.

That is the idea behind the tool.

jmkeil:
I just updated the comparison configuration, revealing further deviations.

jmkeil:
The last update removes deviations obviously caused by OM or SWEET from the result.

steveraysteveray:
Closing for housekeeping reasons...

PR #529, opened 2022-08-01
added new unit gigabase pair (unit of length for DNA molecules)

kitchenprinzessin3880:


steveraysteveray (approved):
I'm inferring that it is not useful to also have the un-prefixed BasePair unit.

PR #531, opened 2022-08-02
added units - RPK (reads per kilobase)

kitchenprinzessin3880:


steveraysteveray (changes requested):
Would you mind also adding the following 2 triples? We're trying to improve on distinguishing units that are dimensionless but not commensurate. This isn't a perfect solution, but at least prevents people from comparing a relative humidity to a temperature ratio... This is discussed [here](https://github.com/qudt/qudt-public-repo/wiki/User-Guide-for-QUDT#6-using-quantity-kinds-with-dimensionless-quantities).

In the current case it is the ratio of two dimensionless numbers.


```
  qudt:qkdvDenominator qkdv:A0E0L0I0M0H0T0D1 ;
  qudt:qkdvNumerator qkdv:A0E0L0I0M0H0T0D1 ;
```

steveraysteveray (changes requested):
Ah, with the new quantity kind, could you add to it:

1. The same two numerator and denominator triples.
2. skos:broader quantitykind:DimensionlessRatio

kitchenprinzessin3880:
> Ah, with the new quantity kind, could you add to it:
> 
>     1. The same two numerator and denominator triples.
> 
>     2. skos:broader quantitykind:DimensionlessRatio

done ;)

steveraysteveray (approved):
Perfect!

PR #535, opened 2022-08-04
Srr small fixes

steveraysteveray:


jhodgesatmb (commented):
For the quantitykind:ParticleCurrent change, the applicable unit says unit:PER-M2-SEC but the dimension vector was changed from qkdv:A0E0L-2I0M0H0T-1D0 to qkdv:A0E0L0I0M0H0T-1D0. How come? Seems like it was correct before.

steveraysteveray:
As it describes at https://qudt.org/vocab/quantitykind/ParticleCurrent.html, ParticleCurrent is the integral over area, rather than being a count per unit area. As such, the dimensionality should be just inverse time.

The applicable unit triple is [auto-generated](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#4-computing-applicable-units-for-a-quantitykind) from the referring units and so unit:PER-M2-SEC will no longer be listed. Your comment does remind me that for ParticleCurrent to show any applicable units, it should include a skos:broader to some inverse time quantitykind, and quantitykind:Frequency looks like the best choice. I will add that.

ralphtq (approved):

PR #541, opened 2022-08-18
Add Dimension Vectors to 10 Quantity Kinds

mgberg:
As a first test of the PR process (both yours and internal to my organization), this includes the linkage of 10 existing Quantity Kinds with the corresponding existing Dimension Vectors.


mgberg:
> Looks good, and made me realize the following query provides a bunch of other values as well. I plan to fill in the remaining blanks revealed by the query, so don't bother with them.
> 
> ```
> SELECT DISTINCT ?qk ?audv
> WHERE {
> ?qk a qudt:QuantityKind .
> FILTER NOT EXISTS {?qk qudt:hasDimensionVector ?dv .}
> ?qk qudt:applicableUnit/qudt:hasDimensionVector ?audv .
> }
> ```

Sounds good, that's basically the same query I had been using.

There's also the possibility that a quantity kind and an applicable unit for that quantity kind have different dimension vectors. I have a SHACL constraint for that (and some other situations) in our system.

Issue #542, opened 2022-08-18
Some xsd:decimal representations are too long for some triple stores


mgberg:
One of the triple stores we're using encodes literals of certain XSD datatypes to enable fast lookup, conversion, and range and order queries on values of those types. Unfortunately, conversion multipliers for a few of the Planck units are so long that they fall out of their spec for accepted values for `xsd:decimal`, and therefore the unit graph cannot be loaded into this triple store without modification.

There are five units which fall in this category: `unit:PlanckArea`, `unit:PlanckCurrentDensity`, `unit:PlanckDensity`, `unit:PlanckPressure`, `unit:PlanckVolume`.

One possible solution is to convert the conversion multiplier for at least these units from `xsd:decimal` to `xsd:double`. However, I'm not sure what the current ruling is on datatypes used for conversion multipliers/offsets. I know that at one point it was decided to make everything `xsd:decimal`s to reduce issues regarding floating-point arithmetic, but it seems that decision was reversed a few months ago.

steveraysteveray:
While we clear this up, I have declared the multipliers for the units you mentioned as xsd:double explicitly. Slipped this fix into #544.

mgberg:
Great! I have confirmed that loading the file as it is in #544 works as intended.

mgberg:
Great! I have confirmed that loading the file as it is in https://github.com/qudt/qudt-public-repo/pull/544 works as intended.

jmkeil:
> One of the triple stores we're using encodes literals of certain XSD datatypes to enable fast lookup, conversion, and range and order queries on values of those types. Unfortunately, conversion multipliers for a few of the Planck units are so long that they fall out of their spec for accepted values for `xsd:decimal`, and therefore the unit graph cannot be loaded into this triple store without modification.

@mgberg Which triplestore implementation causes the problem? [xsd:decimal specifications](https://www.w3.org/TR/xmlschema11-2/#decimal) (including [ minimally conformance requirements](https://www.w3.org/TR/xmlschema11-2/#partial-implementation)) do not limit the length of values. It must be the triplestore, not the values, that doesn't meet the specifications.

jmkeil:
@mgberg Of course, a reasonable design decission. I am curious: Which triplestore software is that? Is the limit configurable or hard coded?

mgberg:
AllegroGraph; their datatype documentation is [here](https://franz.com/agraph/support/documentation/current/datatypes.html).

PR #544, opened 2022-08-18
 Added DimVec where applicableUnit gave answer. Also fixed one error …

steveraysteveray:
…found


jhodgesatmb (commented):
Do all of these units not have dimension vectors? I see that CG is still all caps CENTER_OF_GRAVITY. Are we worried about breaking people's code by not changing it?

Most of my comment is about these conversion multipliers. I never liked the million zeros before or after the significant digits but was overruled. to me these numbers should be exponentials and they once were. Please comment.

steveraysteveray:
I'm OK if you want to change the all upper case to CamelCase.
Regarding the conversion multipliers, I'm also open to revisiting, but keeping in mind the issues that have been raised in #542, #307 and #311.

steveraysteveray:
Agreed to ADD (not change) center of gravity quantity kinds in CamelCase, for backward compatibility.

ralphtq (approved):

PR #546, opened 2022-08-22
Fix units related to Elementary Charge

jmkeil:
Found due to deviation revealed with ABECTO (see #525).

Please also note the duplicate `unit:E` and ` unit:ElementaryCharge` which I was not sure how to handle.


steveraysteveray (approved):
Thanks! Nicely caught.

PR #551, opened 2022-08-31
remove "dcterms:description Unavailable@en" statements from units

jmkeil:


steveraysteveray:
Happy to entertain this idea, but could you talk a little about why you would like them removed? We thought that was helpful to have the statements...

jmkeil:
I think this statements should get removed, as they potentially fool queries or applications regarding the availability of an description. E.g. a (naive) query / a SHACL constraint to find units without descriptions would fail. And applications would need to handle missing descriptions anyhow, as there also exist units without actual or placeholder description.

jmkeil:
`dcterms:description Unavailable@en` states  "I don't know" in natural language, which is not machine understandable. The RDF native way to state "I don't know" is to make no statement.

ralphtq (approved):

PR #553, opened 2022-08-31
fix unit:N-PER-M2 symbol & label (revealed with ABECTO)

jmkeil:


steveraysteveray (approved):
For the time being, we are planning to include both owl:sameAs and qudt:exactMatch to link like Units (and QuantityKinds), but I will create a separate PR to fix existing errors.

jmkeil:
`unit:PA` with the symbol `Pa` and according links already exists. Its just the problem, that `unit:N-PER-M2` features the symbol of the wrong synonym.

PR #556, opened 2022-09-02
Fix some units that are not of quantity kind Currency

mgberg:
There are a few units that were defined as Currency units which (probably aren't):

- Some are currency per time, but that didn't exist yet and am not sure what the appropriate way is to represent the dimensionality of that (is the "dimensionless" exponent 1 or 0?)
- The appropriate dimension vector for PER-H already existed, so that was connected
- The appropriate dimension vector for KiloGM-SEC2 did not exist, so that was added and connected

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -6418,7 +6418,6 @@ unit:FijiDollar
 unit:Flight
   a qudt:Unit ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
-  qudt:hasQuantityKind quantitykind:Currency ;
   qudt:symbol "flight" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Flight"@en ;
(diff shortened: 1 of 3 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -6418,7 +6418,6 @@ unit:FijiDollar
 unit:Flight
   a qudt:Unit ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
```
steveraysteveray, line 6420: Please add qudt:hasQuantityKind quantitykind:Dimensionless ;

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -12749,7 +12747,6 @@ unit:MDOLLAR-PER-FLIGHT
   a qudt:Unit ;
   qudt:expression "\\(\\(M\\$/Flight\\)\\)"^^qudt:LatexString ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
```
steveraysteveray, line 12749: Another legacy from the NASA days! The URI should be changed to MegaDOLLAR-PER-FLIGHT, but the dimensionality is correct - a currency per a dimensionless unit (Flight).

mgberg: Should the URI be changed as part of this PR, or should I just add the quantity kind back?

steveraysteveray (changes requested):
First, thanks so much for your attention to these and subsequent problems!

1. All the qudt:applicableUnit entries are [algorithmically generated](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#4-computing-applicable-units-for-a-quantitykind), so no need to make edits in the quantity kind file - they would just reappear with the next publication. It is the entries in the unit file that drive the algorithm.
2. I have added a few in-line requests for your suggested changes.
3. The currency units are generally a mess and we haven't waded into them to clean them up. First off, they violate our [naming rules](https://github.com/qudt/qudt-public-repo/wiki/Unit-Vocabulary-Submission-Guidelines#qname-naming-rules). So I'm wondering if we should not even try to tweak them, but rather launch an entire "project" to fix them. It has been argued in the past that currency units don't even belong in the units vocabulary, but this has not been settled. All we need is someone willing to take this on!

steveraysteveray (approved):

PR #560, opened 2022-09-07
fix UCUM code of unit:REV derived units (revealed with ABECTO)

jmkeil:
This fixes the UCUM codes of unit:REV derived units replacing `{#}` by `circ`. I am not completely certain, if `circ` is the proper code to use, but it is the only code representing a plane angle equal to 360°. The alternative would be, to remove these `qudt:ucumCode` statements.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -21010,7 +21010,7 @@ unit:REV
   qudt:informativeReference "http://en.wikipedia.org/wiki/Revolution?oldid=494110330"^^xsd:anyURI ;
   qudt:omUnit <http://www.ontology-of-units-of-measure.org/resource/om-2/revolution> ;
   qudt:symbol "rev" ;
-  qudt:ucumCode "{#}"^^qudt:UCUMcs ;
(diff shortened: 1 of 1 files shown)


dr-shorthair:
I'm investigating if this is the best option - see https://github.com/ucum-org/ucum/issues/198

gschadow:
I don't think you want that change.

If you take it you change the kind of quantity of revolutions per minute from frequency to angular velocity.

BTW, I came here through [this discussion thread.](https://github.com/ucum-org/ucum/issues/198)

dr-shorthair:
Hmm. revs-per-minute sounds like an angular velocity to me.

gschadow:
> Hmm. revs-per-minute sounds like an angular velocity to me.

At the very least it's ambiguous. https://en.wikipedia.org/wiki/Revolutions_per_minute, most people think of it as a frequency. 

QUDT says they are a semantic concept system, so I am sure if they defined a symbol RPM, they would have an unambiguous definition for it. 

And if they don't have their own, they had used something like {#}/min, with the {#} annotation probably meaning "number", so they committed to it being a frequency as most people would if you polled them. Even car mechanics, who would adjust timing belts would probably not say "angular velocity" if you asked them.

steveraysteveray:
Given the lack of consensus regarding the use of "circ", I'm closing this PR for now. We can revisit if necessary.

PR #563, opened 2022-09-08
Minor dimensionality and quantity kind fixes

mgberg:
A few fixes I can confidently make:
- `unit:M2-PER-KiloGM` can't be of `quantitykind:MolarAbsorptionCoefficient` - removed it
- `unit:LB-MOL` is not of dimensionality `qkdv:A1E0L0I0M0H0T0D0` - added and set to `qkdv:A1E0L0I0M1H0T0D0`
- `unit:WB-M` has dimension vector `qkdv:A0E-1L3I0M1H0T-2D0` but that dimension vector is not defined - add `qkdv:A0E-1L3I0M1H0T-2D0` and also set it as the dimension vector of `unit:N-M2-PER-A` (which is a `qudt:exactMatch` for `unit:WB-M`)
- `quantitykind:LuminousFlux` is not of dimensionality `qkdv:A0E0L2I0M1H0T-3D0` - set to `qkdv:A0E0L0I1M0H0T0D0`
- `unit:LM` is not of dimensionality `qkdv:A0E0L2I0M1H0T-3D0` - set to `qkdv:A0E0L0I1M0H0T0D0`

Note that the last two are related to #562 .

### DIFF
--- a/vocab/dimensionvectors/VOCAB_QUDT-DIMENSION-VECTORS-v2.1.ttl
+++ b/vocab/dimensionvectors/VOCAB_QUDT-DIMENSION-VECTORS-v2.1.ttl
@@ -410,6 +410,22 @@ qkdv:A0E-1L3I0M0H0T-1D0
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/dimensionvector> ;
   rdfs:label "A0E-1L3I0M0H0T-1D0" ;
 .
+qkdv:A0E-1L3I0M1H0T-2D0
+  a qudt:QuantityKindDimensionVector_ISO ;
+  a qudt:QuantityKindDimensionVector_Imperial ;
+  a qudt:QuantityKindDimensionVector_SI ;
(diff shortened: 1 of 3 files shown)


steveraysteveray (changes requested):
Agreed on M2-PER-KiloGM. A couple of follow-on implications that need cleaning up. Would you mind:
1. Adding qudt:hasQuantityKind quantitykind:MolarAbsorptionCoefficient to unit:M2-PER-MOL
2. Cross referencing quantitykind:MolarAbsorptionCoefficient and quantitykind:MolarAttenuationCoefficient both ways, using owl:sameAs and qudt:exactMatch

I agree with everything else. Thanks especially for the LuminousFlux fix - they are especially slippery! 

Another favor, would you mind updating the informative reference for unit:LM to the newer Wikipedia entry: https://en.wikipedia.org/wiki/Lumen_(unit)

steveraysteveray (approved):

PR #564, opened 2022-09-08
Remove incorrect HP-based units

mgberg:
It appears that the units `unit:HP-PER-M` and `unit:HP-PER-V` are incorrectly defined, or more specifically the URI doesn't follow the naming convention. From the labels/descriptions it seems like they were intended to be `unit:HP_Metric` and `unit:HP_Electric` respectively. Since these already exist, I propose these are deleted.

This PR removes those two units and references to those units.


steveraysteveray (approved):
Great catch, and an embarrassing error!

PR #567, opened 2022-09-12
Correct symbol of Siemens

fennibay:
Hi, I'd like to introduce symbols for some units, we need them for our application.

It's not a comprehensive amendment for all units, and I'm not sure how qudt:symbol is positioned in regards to qudt:ucumCode. Judging from other units, adding these symbols shouldn't hurt. I'd be interested in your feedback.

Many thanks.


steveraysteveray (commented):
Could you provide a little insight into how you are choosing the symbols? For example, what drives your choice of brace, as in "Btu_{it}/h"?

I ask because I'd like to document how symbols are chosen/identified.

fennibay:
Basically, I combined the symbol on `unit:BTU_IT` which is `Btu_{it}` with `/<unit>` in the symbol, as it was done in some other `PER-<unit>` units like `unit:M2-PER-HA` and `unit:M3-PER-HR`

So I can't really state a consistent rule here, but it is consistent with the other examples I found.

dr-shorthair:
UCUM codes are fully consistent and computable - https://ucum.org/ucum 
Why not use those?

fennibay:
Using qudt:ucumCode would be OK from my POV. It is defined on much more units, and apparently has a consistent approach in contrast to qudt:symbol.

But then I would question why qudt:symbol is defined in QUDT schema at all. It is already defined on ~340 units.

steveraysteveray:
Excellent question. The answer is that the intention is for qudt:symbol to provide the Unicode symbol, wherever it exists. Not computable like UCUM, but recognizable to humans. We're not complete on that yet.

steveraysteveray:
So I'm now wondering whether you would be interested in modifying your PR to add the Unicode symbols? For example, many symbols I have not investigated are defined [here](https://unicode-table.com/en/sets/unit-symbols/).

Not a trivial task, I understand. The alternative is to just delete the ones that have Unicode alternatives.

fennibay:
Yes, I could do that. It would take some time of course :-)

I can for instance change cm2 with ㎠. How would the approach be for combined units? Such g/kg and J/K

steveraysteveray:
When you cannot find a pre-defined Unicode symbol, just try to combine the ones that do exist. So, for g/kg, it would be

g/㎏

i.e.
U+0067U+002FU+338F

See https://unicode-table.com/en/ for the entire table of codes.

(If anybody else is more expert in Unicode and knows of a better way, please weigh in!)

🙏 Thanks for your help!

steveraysteveray:
Wow, 1500 Unicode symbols would be a great improvement. Certainly better than our current situation, so please go ahead and contribute! Do note that there's a merge conflict with our current master (probably because of Friday's release?) so please resolve those before the pull request. 
Thanks!

steveraysteveray:
@fennibay, in light of our recent PR #577, we now have a merge conflict with your PR. Would you mind pulling the master branch and resolving the conflicts, assuming you still want the changes you have submitted?

fennibay:
Hi @steveraysteveray, please excuse my late reply. Actually I wanted to add the Unicode symbols, but couldn't find the time for it.

In the meantime #577 provided the symbols for all the units I need (and then some more).

So for this MR, pragmatically I would have one tiny change left: correction of the unit for Siemens, I think it should be uppercase S and not lowercase; because lowecase would indicate seconds, and [Wikipedia page](https://en.wikipedia.org/wiki/Siemens_(unit)) indicates it should be uppercase, as well. // And I happen to be working at @siemens, not that that makes me an expert on this unit ;-)

I'm now updating the title of this PR to reflect the new content.

steveraysteveray (approved):

Issue #568, opened 2022-09-16
dcterms in QUDT inconsistent with DublinCore specification


ektrah:
SCHEMA_QUDT-v2.1.ttl [currently](https://github.com/qudt/qudt-public-repo/blob/4d9466887e68d4e854044b4d568a1066ae42265b/schema/SCHEMA_QUDT-v2.1.ttl) includes some statements about terms from the http://purl.org/dc/terms/ namespace, e.g.:

https://github.com/qudt/qudt-public-repo/blob/4d9466887e68d4e854044b4d568a1066ae42265b/schema/SCHEMA_QUDT-v2.1.ttl#L38-L42

However, the [latest http://purl.org/dc/terms/ specification](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/dublin_core_terms.ttl) defines some of these terms slightly differently, e.g.:

```ttl
dcterms:creator
    dcam:rangeIncludes dcterms:Agent ;
    dcterms:description "Recommended practice is to identify the creator with a URI.  If this is not possible or feasible, a literal value that identifies the creator may be provided."@en ;
    dcterms:issued "2008-01-14"^^<http://www.w3.org/2001/XMLSchema#date> ;
    a rdf:Property ;
    rdfs:comment "An entity responsible for making the resource."@en ;
    rdfs:isDefinedBy <http://purl.org/dc/terms/> ;
    rdfs:label "Creator"@en ;
    rdfs:subPropertyOf <http://purl.org/dc/elements/1.1/creator>, dcterms:contributor ;
    owl:equivalentProperty <http://xmlns.com/foaf/0.1/maker> .
```

In this case, the `rdfs:range` statement from QUDT turns every instance of an `dcterms:Agent` referenced in a `dcterms:creator` statement implicitly into an instance of `xsd:string`, which makes some tools unhappy.

Could the terms from DublinCore be just referenced instead of being duplicated into QUDT?

steveraysteveray:
I understand your concern, and will remove the rdfs:range triple for both dcterms:creator and dcterms:contributor. Hopefully that will solve your problem. Stay tuned for a pull request, and a new Release probably this week.

PR #572, opened 2022-09-19
 Removed range restriction of xsd:string for dcterms:creator and dcte…

steveraysteveray:
…rms:contributor


steveraysteveray:
This closes #568.

ralphtq (approved):

PR #581, opened 2022-10-14
Fix Siemens units

paroga:


steveraysteveray (changes requested):
Nice catch, thanks! 
Would you mind also updating the dimension vector for quantitykind:Admittance?

steveraysteveray (approved):

Issue #593, opened 2022-10-25
quantitykind:VolumePerArea referenced but not defined


fkleedorfer:
```
unit:M3-PER-HA qudt:hasQuantityKind quantitykind:VolumePerArea 
```
but there are no triples with `quantitykind:VolumePerArea` in the subject position.

`quantitykind:VolumePerArea` is new in v2.1.20, which probably means that we should add its definition. I'll start work on a PR with what I can come up with myself, but I'll probably need a bit of help to fill in all the properties.

fkleedorfer:
Digging. `quantitykind:VolumePerArea` was introduced in #571

fkleedorfer:
I am not sure about the quantitykind. `unit:M3-PER-HA`, which introduces `quantitykind:VolumePerArea`, has this vector:
```
unit:M3-PER-HA  qudt:hasDimensionVector qkdv:A0E0L1I0M0H0T0D0 .
```
... which looks wrong to me. We are using the same dimension in nominator and denominator, but with different exponents. Here, the idea seems to be that they can be reduced as `L^3/L^2 = L^1`. I believe this is the only unit that has unequal positive and negative exponents for the same dimension. In other cases, where both exponents are the same, such as `unit:M3-PER-M3`, we instead say that it's just a ratio (`qkdv:A0E0L0I0M0H0T0D1`). 

This is an interesting case, and I see three options:
* use `qkdv:A0E0L1I0M0H0T0D0` and say, `unit:M3-PER-HA` can be converted into any length.
* use `qkdv:A0E0L0I0M0H0T0D1` and say it is a ratio, theoretically convertible into any other ratio, but that hasn't bothered anyone so far
* try something new and use `qkdv:A0E0L3-2I0M0H0T0D0`, making it a very narrow category of ratios, but we could convert between, say `unit:M3-PER-HA` and a (hypothetical) `unit:MilliM3-PER-MilliM2`. Or use `qkdv:A0E0L3-2I0M0H0T0D1` to still mark it as a ratio.

Food for thought, but in the PR, I'll go with the vector used by `unit:M3-PER-HA`

steveraysteveray:
I believe your first option is best, but also including the numerator and denominator dimensions.

[...]

PR #594, opened 2022-10-25
Add missing quantitykind:VolumePerArea

fkleedorfer:
This quantitykind was referenced by unit:M3-PER-HA but not defined(#593). 

This PR addresses this problem.

As [stated in the issue](https://github.com/qudt/qudt-public-repo/issues/593#issuecomment-1290128388), this PR uses the dimension vector `qkdv:A0E0L1I0M0H0T0D0`, but another one might be better.


steveraysteveray (changes requested):
I'll comment further in issue #593, but support what you have done so far.

One further request is that you include the qudt:qkdvNumerator and qudt:qkdvDenominator triples (with L3 and L2 dimensions).

steveraysteveray:
We have discussed this further, and now agree that it's a slippery slope to use the numerator and denominator for anything other than dimensionless ratios, so we're going ahead and approving this PR.

steveraysteveray (approved):

Issue #327, opened 2020-12-23
Move currency units into separate file/graph


dr-shorthair:
(Second try)

Currency units are interesting, but rather a different thing to the primary topic of this work. Suggest moving them all into a separate graph, so that the main catalogue is more cleanly about observations quantities.

jhodgesatmb:
QUDT is not OQUDT. It is also not QQUDT (quantifiable quantities, units, dimensions, and data types). It is about all quantities, etc. I can see moving the currency units to another graph but not for that reason.

Jack

nicholascar:
I support moving currency into its own graph, not because I don't like the currency info, but I think it would be more easily used in isolation if separated. It's sort of the inverse to the starting issue but if QUDT wants people to use QUDT for currency things, currencies should be _called out_ by placing them in a separate graph.

Surely of all the things QUDT does, currency is pretty easy!

dr-shorthair:
> Surely of all the things QUDT does, currency is pretty easy!

Except that conversion multipliers are time-dependent. 
In my opinion that puts them into a very different class to all other units. 
I would suggest that the QUDT model needs to be specifically modified to support that fact, and absent that there is very little utility in having currencies as part of the catalogue.

nicholascar:
I assumed there were no conversion multipliers for currency in QUDT due to the time dependence and I didn’t even bother to check this thinking it was obvious...

It would be a massive undertaking to maintain a current set of currency multipliers. Not feasible.

dr-shorthair:
... which is why this whole class of units should be managed differently.

steveraysteveray:
Turns out that the scripts that publish the web pages introduce some non-trivial issues when I started editing them to publish currencies separately. I still have the branch in our internal repo, so I could go back to take another run (should be doable eventually), but I don't think we have consensus yet on the Board.

PR #604, opened 2022-11-11
Add units and quantitykinds and maps QUDT to IFC using sssom

fkleedorfer:
Addresses #599 

* Adds units required for aligning with IFC
* Adds quantitykinds required for aligning with IFC
* Adds mappings between qudt units and IfcSIUnitName classes
* Adds mappings between qudt quantitykinds and Ifc MeasureValue and IfcDerivedMeasureValue classes

I've added a minmalistic SSSOM mapping instead of none at all because I think it is easier to cross-check the additions in the light of which IFC classes they are being mapped to. Also, I thought it is beneficial to have the mappings out there so they can improve over time.


steveraysteveray (commented):
The QUDT team will need to re-serialize the vocabulary files after this merge, so that additions get combined with existing triples. Otherwise, it looks good to me.

fkleedorfer:
Sorry I found one more quantitykind/unit combination relevant for IFC (see latest commit [17285](https://github.com/qudt/qudt-public-repo/pull/604/commits/17285822003c61c89689a66c6ce53e42320cc76d)). They are used in models but not explicitly defined in the standard - hence, no additional mappings for these (it's a "user defined" unit according to IFC).

As we get on with our project, we expect encountering units/quantitykinds that are relevant in the domain and not yet part of QUDT. Probably a handful in the near-to-medium future, as we work with increasingly realistic models, and then, hopefully, fewer and fewer.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -18676,3 +18676,11 @@ quantitykind:WarpingMoment
     rdfs:label "Warping Moment"@en ;
 .
 
+quantitykind:FrictionLoss
+    a qudt:QuantityKind ;
+    qudt:hasDimensionVector qkdv:A0E0L-2I0M1H0T-2D0;
```
steveraysteveray, line 18681: The dimension seems wrong. If you really mean "power", then the dimension should be A0E0L2I0M1H0T-3D0. If you mean energy, then it would be A0E0L2I0M1H0T-2D0

steveraysteveray: You might also want to include a skos:broader statement.

steveraysteveray:
@fkleedorfer, your PR caused us to think ahead a little about how to manage community contributions in general. Accordingly, we just made some changes to the community folder (see the [README](https://github.com/qudt/qudt-public-repo/blob/master/community/README.md)). I would ask that you pull the master branch, resolve any conflicts, and put your new files in the community/mappings/SSSOM/IFC folder, along with a README.md in the same folder.

Sorry for changing the landscape under you, but we think this will help in the longer run. We like what you have done so far!

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -18410,3 +18410,277 @@ vaem:GMD_QUDT-QUANTITY-KINDS-ALL
...
+    qudt:informativeReference "https://en.wikipedia.org/wiki/Acid"^^xsd:anyURI ;
+    qudt:informativeReference "https://en.wikipedia.org/wiki/PH"^^xsd:anyURI ;
+    rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
+    rdfs:label "PH"@en ;
+.
+
+quantitykind:Acidity
+    a qudt:QuantityKind ;
+    qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
+    qudt:plainTextDescription "Chemicals or substances having a pH less than 7 are said to be acidic; lower pH means higher acitidy."@en ;
```
steveraysteveray, line 18624: Typo - acitidy

steveraysteveray (changes requested):

fkleedorfer:
No problem, I'll rebase and provide a readme.

Another thing is popping up during alignment:
IFC also has currency units, see
https://standards.buildingsmart.org/IFC/DEV/IFC4_3/RC1/HTML/link/ifcmonetaryunit.htm which states that the currency label must conform to [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217). 

So, for mapping currencies these between IFC and QUDT, it would be necessary to have the ISO4217 currency code available in the units. I see that they are mapped to dbpedia, which in turn has `dbp:isoCode`, `dbp:isoNumber` and `dbp:isoExponent`. Qudt, on the other hand, has only `qudt:currencyExponent`.

I could add this to this PR or make another one, but this would be the suggestion:
- [ ] define qudt:currencyCode, pointing to ISO 4217, the wikipedia entry (which is more accessible), and dbp:isoCode
- [ ] define qudt:currencyNumber, pointing to ISO 4217, the wikipedia entry (which is more accessible), and dbp:isoNumber
- [ ] add both to all currencies, e.g.:
```
unit:Euro 
    qudt:currencyCode "EUR" ;
    qudt:currencyNumber 978 ;
```

What do you think?

steveraysteveray:
Let's do one thing at a time and finish up the IFC work in this PR.

I recommend you bring up the currency issue as a separate Issue. Note that the treatment of currency has been discussed before (notably #327), but your suggestion sounds like a good one to me whether currency is separated or not.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
fkleedorfer: The unit associated with the name 'friction loss' in the IFC model I was working on was kg / (m²s²), that's where I took the dimensions from. In the wikipedia article, https://en.wikipedia.org/wiki/Friction_loss, this unit is referred to as 'friction loss per unit length'. I had to add that unit to the units file as well. 

I took the description from the wikipedia article, which covers friction loss in a number of different fields. I may have gotten the wrong one, given that the unit kg/(m²s²) appears in the section on 'Calculating friction loss for air in a duct'. Here are the 3 definitions of 'friction loss' (I used the second, but the first seems better suited):
 
> The term friction loss (or frictional loss) has a number of different meanings, depending on its context.
>
> In [fluid flow](https://en.wikipedia.org/wiki/Fluid_flow) it is the [head loss](https://en.wikipedia.org/wiki/Head_loss) that  occurs in a containment such as a pipe or duct due to the effect of the fluid's [viscosity](https://en.wikipedia.org/wiki/Viscosity) near the surface of the containment.[[1]](https://en.wikipedia.org/wiki/Friction_loss#cite_note-Munson-1)

> In mechanical systems such as [internal combustion engines](https://en.wikipedia.org/wiki/Internal_combustion_engine), the term refers to the power lost in overcoming the [friction](https://en.wikipedia.org/wiki/Friction) between two moving surfaces. 

> In economics, frictional loss is natural and irrecoverable loss in a transaction or the cost(s) of doing business too small to account for. Contrast with [tret](https://en.wikipedia.org/wiki/Trett) in shipping, which made a general allowance for otherwise unaccounted for factors.

How about we rename the quantity kind to `qk:FrictionLossPerLength`, use the first definition and keep the dimension vector? If there is a good candidate for a broader candidate, I did not find it. Help appreciated!

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -18410,3 +18410,277 @@ vaem:GMD_QUDT-QUANTITY-KINDS-ALL
...
+  qudt:baseUnitDimensions "\\(L^3/M\\)"^^qudt:LatexString ;
+  qudt:hasDimensionVector qkdv:A0E0L3I0M-1H0T0D0 ;
+  qudt:informativeReference "https://bigladdersoftware.com/epx/docs/8-4/engineering-reference/effective-moisture-penetration-depth-empd.html#empd-nomenclature"^^xsd:anyURI ;
+  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
+  rdfs:label "Isothermal Moisture Capacity"@en ;
+  skos:broader quantitykind:SpecificVolume .
+
+quantitykind:MoistureDiffusivity # TODO unsure about the unit/dimensionality. See https://github.com/buildingSMART/IFC4.3.x-development/issues/547
+  a qudt:QuantityKind ;
+  skos:broader quantitykind:VolumeFlowRate ;
```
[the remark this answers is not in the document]
fkleedorfer: ok

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
fkleedorfer: ok

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Actually, how about we make the URI qk:PressureLossPerLength, which is actually more correct, and then in the description we can say that this is sometimes referred to as "Friction Loss?"

My reasoning is that "friction" is actually a force, while "pressure" is a force per area. The Wikipedia article you mention shows the table as a delta (Pressure) / Length.

fkleedorfer: ok, done.

steveraysteveray (approved):

cmungall:
Thanks for doing this!

Note this won't pass validation as it doesn't include a https://w3id.org/sssom/mapping_justification field, indicating whether this was manual or lexical etc

fkleedorfer:
It was all painstakingly manual. What would need to be added exactly?

PR #606, opened 2022-11-14
Minor fixes to a few labels, symbols, and a dimension vector

mgberg:
I found a few minor issues while going through some internal updates today:
- Fixed dimension vector of `quantitykind:MolarAbsorptionCoefficient`
- Fixed a couple symbols and labels for consistency
- Removed a couple symbols for 2 units that had been removed.


steveraysteveray (approved):
Thanks for catching the dimension vector. I may have been the guilty party, thinking mole but using M instead of A!

PR #618, opened 2022-12-05
 Fixed a few errors in recent additions

steveraysteveray:


steveraysteveray:
Precisely correct!
We are currently implementing a CI suite in GitHub - not quite ready yet, but it is lurking in this repo. Planning to invoke it upon pull requests, and manually by contributors.

PR #622, opened 2023-01-10
Changes to QuantityKinds that have no dimension vectors

steveraysteveray:


vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -11134,23 +11164,28 @@ quantitykind:MechanicalEnergy
 .
 quantitykind:MechanicalImpedance
   a qudt:QuantityKind ;
+  qudt:hasDimensionVector qkdv:A0E-2L2I0M1H0T-3D0 ;
```
steveraysteveray, line 11167: I'm wondering why it isn't Mass/Time.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -11134,23 +11164,28 @@ quantitykind:MechanicalEnergy
...
 quantitykind:MechanicalMobility
   a qudt:QuantityKind ;
   qudt:applicableUnit unit:MOHM ;
+  qudt:hasDimensionVector qkdv:A0E0L0I0M-1H0T1D0 ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "Mechanical Mobility"@en ;
 .
 quantitykind:MechanicalSurfaceImpedance
   a qudt:QuantityKind ;
+  qudt:hasDimensionVector qkdv:A0E-2L2I0M1H0T-3D0 ;
```
steveraysteveray, line 11181: I'm finding Acoustic impedance to have dimension Mass/Length**3 . time, so this would yield Mass/Time.
But I could be wrong.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Not a narrower version of quantitykind:Impedance, which is electrical.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -16068,12 +16128,14 @@ quantitykind:SpecificOpticalRotatoryPower
 quantitykind:SpecificThrust
   a qudt:QuantityKind ;
   qudt:dbpediaMatch "http://dbpedia.org/resource/Specific_thrust"^^xsd:anyURI ;
+  qudt:hasDimensionVector qkdv:A0E0L1I0M1H0T-2D0 ;
```
steveraysteveray, line 16131: I'm thinking Length/Time

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -16068,12 +16128,14 @@ quantitykind:SpecificOpticalRotatoryPower
...
   qudt:id "Q-160-100" ;
   qudt:informativeReference "http://en.wikipedia.org/wiki/Specific_thrust"^^xsd:anyURI ;
   qudt:plainTextDescription "Specific impulse (usually abbreviated Isp) is a way to describe the efficiency of rocket and jet engines. It represents the force with respect to the amount of propellant used per unit time.[1] If the \"amount\" of propellant is given in terms of mass (such as kilograms), then specific impulse has units of velocity. If it is given in terms of Earth-weight (such as kiloponds), then specific impulse has units of time. The conversion constant between the two versions of specific impulse is g. The higher the specific impulse, the lower the propellant flow rate required for a given thrust, and in the case of a rocket the less propellant is needed for a given delta-v per the Tsiolkovsky rocket equation." ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "Specific thrust"@en ;
   rdfs:seeAlso quantitykind:SpecificImpulse ;
+  skos:broader quantitykind:Thrust ;
```
steveraysteveray, line 16138: Not valid dimensionally.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -16127,6 +16189,7 @@ quantitykind:SpectralLuminousEfficiency
   qudt:systemDerivedQuantityKind qudt:SOQ_SI ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "Spectral Luminous Efficiency"@en ;
+  skos:broader quantitykind:LuminousEfficacy ;
```
steveraysteveray, line 16192: Efficiency would be a dimensionless percentage; efficacy is, for example lm/W which is not dimensionless, so one cannot be skos:broader than the other.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -17903,6 +17980,7 @@ quantitykind:Turns
 .
 quantitykind:UniversalGasConstant
```
steveraysteveray, line 17981: Suggest we move this to the PhysicalConstant graph

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -6747,31 +6752,39 @@ quantitykind:ForcePerLength
 .
 quantitykind:FractionalMass_Stage1
   a qudt:QuantityKind ;
+  qudt:hasDimensionVector qkdv:A0E0L0I0M1H0T0D0 ;
```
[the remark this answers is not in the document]
jhodgesatmb: The label said fractional mass allocated to Stage 1 <2, 3, 1-3> so I took that to mean an amount and not a fraction. Ralph would know I suppose. I found no web reference.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Then that is my mistake. I assumed that it was the same units as Impedance and took the dimension vector from Impedance.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: There was an applicable unit of MOHM so I took the dimension vector from it, which is qkdv:A0E0L0I0M-1H0T1D0.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Again, I assumed that specific thrust was a type of thrust w/o thinking, but Wikipedia says it is force/mass flow rate, which agrees with what you said.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: I never made any changes for LuminousEfficacy as far as I can tell, and in my spreadsheet I have a value for DV but the row is red so nothing there is valid. I did take from the plain text description this: "the ratio of luminous flux (in lumens) to power (usually measured in watts)" and tried to figure out the DV from there, but there was no such DV in the vocabulary so I mentioned it to you.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Agreed, Mass/Time.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Take another look.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Agreed.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Agreed to remove the skos:broader.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Need to fix LM-PER-W in units:
Remove ref to quantitykind:SpectralLuminousEfficiency
Make DV: L-2, I1, M-1, T3

, and make quantitykind:SpectralLuminousEfficiency dimensionless

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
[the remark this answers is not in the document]
jhodgesatmb: Until I know what 'retire' means I have changed the DVs to Dimensionless

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Fixed

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Does this still have a skos:broader of quantitykind:Impedence? If so then the dimension vectors should be the same, right?

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: fixed, and removed the skos:broader of quantitykind:Thrust

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Fixed

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Did you know that we have the following (nonsensical) dimension vector: A0E0L-2I1M-1H0T3D1

I believe that this 'should be': A0E0L-2I1M-1H0T3D0, which is the DV we need for LM-PER-W. The above is not used anywhere I can see.

jhodgesatmb: Added new dimension vector and associated it with LM-PER-W. Made other changes.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: In performing this task I found out that we already have it: constant:MolarGasConstant. I made the change anyway, but we should decide which if not both we want to keep.

steveraysteveray: Molar has qkdv:A-1E0L2I0M1H-1T-2D0 and value of 8.314472 (should be 8.31446261815324 according to Wikipedia). Universal should have the same DV but doesn't. I suggest we keep them both, fix the Universal DV reference, and have it point to the existing constant:Value_MolarGasConstant. Also add a qudt:exactMatch relation between the two.

...which makes me notice that we still have some skos:exactMatch relations sprinkled in our vocabs that we should change to qudt:exactMatch.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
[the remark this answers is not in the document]
steveraysteveray: I'm pretty sure nobody is using these, so I think we should just delete them. They are always retrievable from earlier releases.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: MechanicalImpedance = AcousticImpedance x Area, so those two seem correct.

...and to answer your question, Impedance is an electrical concept, so no, it is not skos:broader than any of the mechanical analogs. Wikipedia: "There is a [close analogy](https://en.wikipedia.org/wiki/Mechanical%E2%80%93electrical_analogies#Impedance_analogies) with [electrical impedance](https://en.wikipedia.org/wiki/Electrical_impedance), which measures the opposition that a system presents to the [electric current](https://en.wikipedia.org/wiki/Electric_current) resulting from a [voltage](https://en.wikipedia.org/wiki/Voltage) applied to the system."

Which leaves the hard one, MechanicalSurfaceImpedance. [This source](https://apps.dtic.mil/sti/pdfs/ADA315595.pdf) implies it is dimensionally equivalent to pressure/velocity, which by my calculations is Mass/(time x length**2). [Other sources](https://www.wikidata.org/wiki/Q6421317) state it is just force/speed (i.e. not force per unit area/speed), which is mass/time, and is synonymous with MechanicalImpedance. Yet another source, listed [here](https://www.sciencedirect.com/topics/engineering/mechanical-impedance) defines mechanical impedance Z as the square root of (pressure/velocity)!

 Having run out of any other sources, I suggest we define it dimensionally as pressure/speed with DV of Mass/(time x length**2) and point to the [first reference](https://apps.dtic.mil/sti/pdfs/ADA315595.pdf). Perhaps we should include in the comment that we are seeking resolution.

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: I have made these changes. I ran a query looking for skos:exactMatch on all things below qudt:Concept and found several but they are all quantity kinds. Do you want them fixed?

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: Fixed

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
jhodgesatmb: I changed the value to the one on wikipedia.

jhodgesatmb: I am changing the skos:exactMatch values to qudt:exactMatch. I have found one, for ElementaryCharge, that points to quantitykind:AtomicUnitOfCharge, that appears to not exist. When finished I will look into that further.

jhodgesatmb: There is no quantitykind:AtomicUnitOfCharge in the vocabulary.

steveraysteveray: ...but there is a constant:AtomicUnitOfCharge, which makes sense.

There is also constant:ElementaryCharge
I suggest those both should live only in the constants file, related by qudt:exactMatch.
Currently they each have separate constant:Value_... instances, so probably not worth having them both point to the same one.

In addition, I think it is valid to keep the unit:ElementaryCharge instance, because it is used as a unit of measure. (Recall my point that any constant can be used as a unit of measure, but only some are actually used that way).

We should remove references to quantitykind:AtomicUnitOfCharge

jhodgesatmb: I will try to change from qk to c for AtomicUnitOfCharge, though when I load the qk graph the constants are not referenceable so I'll have to do it by editing the text version.

jhodgesatmb: I haven't done anything but change from qk:AUOC to constant:AUOC on qudt:exactMatch and removed the skos:exactMatch because the other comments you made are outside the scope of this PR (as a matter of fact we have ventured pretty far off that PR). My query is no longer finding any skos:exactMatch occurrences.

jhodgesatmb:
I believe that I have addressed all the comments

steveraysteveray:
@jhodgesatmb, I made a few tweaks, plus realized that many of skos:exactMatch triples that you converted to qudt:exactMatch triples (as we agreed) should never have been any kind of exactMatch, since they were references to HTML pages. I changed those predicates to qudt:informativeReference.

If you are good with everything at this point, I think we can merge. See what you think.

jhodgesatmb:
They didn’t all seem like exact matches but I was doing as wished. I am glad you caught it. Yes, wrap it!Jack Hodges, Ph.D.Arbor StudiosOn Jan 13, 2023, at 7:28 AM, steveraysteveray ***@***.***> wrote:﻿
@jhodgesatmb, I made a few tweaks, plus realized that many of skos:exactMatch triples that you converted to qudt:exactMatch triples (as we agreed) should never have been any kind of exactMatch, since they were references to HTML pages. I changed those predicates to qudt:informativeReference.
If you are good with everything at this point, I think we can merge. See what you think.

PR #633, opened 2023-01-26
Add um/s^2, update acceleration to linear acceleration QK for m/s^2 & cm/^2

nicholascar:
* adds the derives unit micrometre per square second
    * copies centimetre per square unit in form
* replaces Acceleration with LinearAcceleration for m/s^2 & cm/^2


steveraysteveray (changes requested):
quantitykind:Acceleration and quantitykind:LinearAcceleration are related by qudt:exactMatch. This is because we felt the "common" interpretation of acceleration is the linear variety. This is one example where we didn't want our vocab to be cumbersome to use by being too pedantic. (We talk about this in the wiki [here](https://github.com/qudt/qudt-public-repo/wiki/Unit-Vocabulary-Submission-Guidelines#synonyms)). So I suggest we refer to both quantitykind:Acceleration and quantitykind:LinearAcceleration from the relevant units so that people will be sure to get what they need.
 
Would you mind updating all 11 (plus your new one) such units?

steveraysteveray:
The remaining piece of this, adding MicroM-PER-SEC2, was done in #1257, so this PR is being closed.

PR #644, opened 2023-02-07
Fix multiplier of unit:LB-PER_IN2

fkleedorfer:


fkleedorfer:
I found 13 units with a multiplier of 0.0. in many cases, I think the best option is to just remove the triple. I'll see which ones I can correct.

steveraysteveray (approved):

PR #646, opened 2023-02-09
 Updated mortality & morbidity units

steveraysteveray:


jhodgesatmb (approved):
Personally I do not agree with this unless there are Units for DEATHS and INDIVS (please note plural forms for counting units).

PR #630, opened 2023-01-20
Extend the QUDT/IFC mapping

fkleedorfer:
Addresses #626 (the missing quantitykinds mentioned in [this comment](https://github.com/qudt/qudt-public-repo/issues/626#issuecomment-1385614885))

* fixes some whitespace errors in the qudt-ifc mapping tsv file
* adds three quantitykinds that are present in ifc:
  - qk:NonNegativeLength
  - qk:PositiveLength
  - qk:PositivePlaneAngle

PR #647, opened 2023-02-15
Deprecated some xyzPercentage qks in favor of xyzRatio

steveraysteveray:


jhodgesatmb (commented):
Although I do not recall a discussion changing percentages to ratios I am not sure I see them as identical. But I could approve this change with some discussion.

What I do not understand are items such as NonNegativeLength, PositiveLength, PositivePlaneAngle, etc. which appear to me to be Quantities rather than QuantityKinds.

steveraysteveray:
Happy to discuss. Note Rule 10 of our [quantity kind naming rules](https://github.com/qudt/qudt-public-repo/wiki/QuantityKind-Vocabulary-Submission-Guidelines#qname-naming-rules). Also, the percentage names are not gone - just marked as owl:deprecated.

Regarding the other items you mentioned, they were not added in this PR, just re-serialized. They were discussed in #626 and merged in #630.

jhodgesatmb (approved):

PR #658, opened 2023-03-10
Add currency symbols for a subset of QUDT currency units.

### LINKED ISSUE #657: Missing currency symbols
We like to show currency symbols, I think it makes sense to manage them in QUDT, at least for the more widely used currencies.
We've already added them to qudtlib: https://github.com/qudtlib/qudtlib-js/pull/20

PR coming up.

fkleedorfer:
There are edge cases. Many currencies have multiple symbols, e.g. Mexican Peso, which can be '$' or 'Mex$'. In this commit, there is only one symbol per currency unit. The decision was made based on the first sentence of the abstract of the respective Wikipedia article (e.g. https://en.wikipedia.org/wiki/Mexican_peso, which states 'The Mexican peso (symbol: $; code: MXN) is the currency of Mexico.')

Actual work done by @clemensjur - thanks!

Fixes #657


steveraysteveray (approved):
Looks good. The US Dollar will now have 2 symbols because we already had US$ as a symbol. That's the only overlap.

fkleedorfer:
Oh, I hadn't seen the US$ - I would have removed it, as I thought that a unit should have only one symbol (which, I believe, is an assumption qudtlib makes at the moment)

PR #667, opened 2023-03-31
resolves #666 Adding Therms (US)/Hour

### LINKED ISSUE #666: Missing Unit: Therm (US) / hour
qudt already has Therm (https://qudt.org/vocab/unit/THM_US) & https://qudt.org/vocab/unit/BTU_IT-PER-HR, so I think it follows suit for therms-per-hour to exist as well.

I went ahead and put together a PR for this as well https://github.com/qudt/qudt-public-repo/pull/667 (since it appears I failed to do something to get these to link to eachother)

ohshazbot:
* Adds therms/hour

This was effectively done by mashing together BTU_IT-PER-HOUR with THM_US since I'm not that intimately familiar with everything here. Please let me know if I missed anything and I'll correct it.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -25338,6 +25338,24 @@ unit:THM_US
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Therm US"@en ;
 .
+unit:THM_US-PER-HR
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (changes requested):
Pretty close! You need to divide the conversionMultiplier by 3600, which by my calculation gives 29300.1111
Then, it would be nice if you found a description of the unit as a unit of power, rather than just describing the Therm.
Finally, we are in the process of deprecating qudt:definedUnitOfSystem in favor of qudt:applicableSystem (this will be in the next Release), so you don't need qudt:definedUnitOfSystem.

Regarding the UNECE code, 2I is for a BTU/hr, while a US Therm/hr is 100,000 BTU/hr, so I think we need to find a different code, yes?

PR #669, opened 2023-04-04
Srr deprecation rule

steveraysteveray:


schema/shacl/SCHEMA_QUDT_NoOWL-v2.1.ttl
```diff
@@ -2302,16 +2304,10 @@ qudt:allowedUnitOfSystem
...
   rdfs:isDefinedBy <http://qudt.org/2.1/schema/shacl/qudt> ;
   rdfs:label "allowed unit of system" ;
   rdfs:subPropertyOf qudt:isUnitOfSystem ;
-.
-qudt:applicableSystem
-  a rdf:Property ;
-  dcterms:description "This property relates a unit of measure with a unit system that may or may not define the unit, but within which the unit is compatible."^^rdf:HTML ;
-  rdfs:isDefinedBy <http://qudt.org/2.1/schema/qudt> ;
-  rdfs:label "applicable system" ;
+  owl:deprecated true ;
```
ralphtq, line 2310: Use qudt:deprecated - this needs to be added as a property. Also will need to be on a property shape on qudt:Concept.

ralphtq (changes requested):

schema/shacl/SCHEMA_QUDT_NoOWL-v2.1.ttl
steveraysteveray: OK, I think this is all fixed up and ready for review. There are quite a few changes (see the commits) but I have tested the whole set with my committee work data files that import QUDT and it behaves as expected:
image

ralphtq (approved):

PR #681, opened 2023-04-18
Fix the namespace IRI of prefix: in SOU

ektrah:


steveraysteveray:
Thanks. I'll push back on the versioned prefixes such as http://qudt.org/2.1/vocab/prefix/Atto. The versionless URI is defined: [http://qudt.org/vocab/prefix/Atto](http://qudt.org/vocab/prefix/Atto). We only use version numbers for the graphs, not for individuals.

I'll look into the others.

ektrah:
I believe all the `http://qudt.org/vocab/type/` and `http://qudt.org/schema/type/` ones should be `http://qudt.org/schema/qudt/`. I'll prepare a PR. I'm not sure about the rest.

steveraysteveray:
Yes, take a look at the PR I just submitted to see if you agree.

Issue #289, opened 2020-11-23
overlap between `qudt` and `datatypes`


VladimirAlexiev:
There is significant overlap between the `qudt` and `datatypes` schemas:
- 784 `datatypes` terms (classes/props)
- 286 `qudt` terms (classes/props)
- 87 overlapping terms (see below)

This creates some complexity and potential confusion. It would be nice to move these terms to only one of the schemas (I guess `datatypes`).

Here are the overlaps:
```
qudt:abbreviation
qudt:allowedPattern
qudt:ansiSQLName
qudt:basis
qudt:BigEndian
qudt:BitEncodingType
qudt:bitOrder
qudt:bits
qudt:BooleanEncoding
qudt:BooleanEncodingType
qudt:bounded
qudt:ByteEncodingType
qudt:byteOrder
qudt:bytes
qudt:cardinality
qudt:CardinalityType
qudt:CharEncoding
qudt:CharEncodingType
qudt:cName
qudt:Concept
qudt:CT_COUNTABLY-INFINITE
qudt:CT_FINITE
qudt:CT_UNCOUNTABLE
qudt:Datatype
qudt:DateTimeStringEncodingType
qudt:description
qudt:DoublePrecisionEncoding
qudt:elementKind
qudt:elementType
qudt:encoding
qudt:Encoding
qudt:EndianType
qudt:EnumeratedValue
qudt:Enumeration
qudt:EnumerationScale
qudt:Figure
qudt:FloatingPointEncodingType
qudt:id
qudt:IEEE754_1985RealEncoding
qudt:IntegerEncodingType
qudt:ISO8601-UTCDateTime-BasicFormat
qudt:javaName
qudt:jsName
qudt:length
qudt:LittleEndian
qudt:LongUnsignedIntegerEncoding
qudt:lowerBound
qudt:MathFunctionType
qudt:matlabName
qudt:maxExclusive
qudt:maxInclusive
qudt:microsoftSQLServerName
qudt:minExclusive
qudt:minInclusive
qudt:mySQLName
qudt:OctetEncoding
qudt:odbcName
qudt:oleDBName
qudt:oracleSQLName
qudt:orderedType
qudt:OrderedType
qudt:Organization
qudt:PartiallyOrdered
qudt:protocolBuffersName
qudt:pythonName
qudt:rdfsDatatype
qudt:Rule
qudt:ScalarDatatype
qudt:ShortSignedIntegerEncoding
[...]

PR #684, opened 2023-04-18
Fix the namespace IRI of qudt.type:

ektrah:
Based on the discussion in https://github.com/qudt/qudt-public-repo/pull/681

@steveraysteveray, please take a look. I'm not sure if I did this right.

It might be easier to review the commits individually than the whole PR.

I didn't try to fix [VOCAB_QUDT-DATATYPES--sDEPRECATED-v2.1.ttl](https://github.com/qudt/qudt-public-repo/blob/main/vocab/types/TBR/VOCAB_QUDT-DATATYPES--sDEPRECATED-v2.1.ttl) since it's marked as deprecated.


ektrah:
Related to https://github.com/qudt/qudt-public-repo/issues/289

steveraysteveray:
The "T" in QUDT is intended to support a robust type model (not unlike what a strongly typed compiler might have), distinct from the values of Units or QuantityKinds. We haven't revisited our earlier work on this, but we want to keep the existing type schema as is for now, distinct from the concepts in the QUDT schema itself. We want to support concepts such as arrays, vectors, tensors, time series, and even arrays of types, but there are many distinctions involved. Therefore, we would rather hold off on any edits to the type schema/prefixes until that time.

jhodgesatmb:
What is the motivation for this request? The datatypes graph is an invaluable component of QUDT for engineering. Do you want a cleaner prefix/namespace or do you want other changes, and why?

ektrah:
As mentioned in https://github.com/qudt/qudt-public-repo/pull/681#issuecomment-1513163362, I was looking at the IRIs in the files of the QUDT v2.1.26 release that show up as predicates and/or objects but never as subjects. This is often an indication that something in the ontology is different than intended (e.g. because a typo has crept in). If after review it is clear that the IRI is correct and it never appears as a subject on purpose, nothing needs to be done.

When it comes to data types, I noticed that QUDT v2.1.26 uses e.g. both `<http://qudt.org/schema/qudt/Double_X>` and `<http://qudt.org/schema/type#Double_X>` , which (at least at first glance) seems like an obvious inconsistency. So I created this PR (just like all the other PRs) in the hope that this inconsistency could be solved quickly. If it doesn't, feel free to close 🙂

Issue #481, opened 2022-03-29
Inconsistent use of qudt prefix


Mttbnchtt:
Most files defines the qudt prefix as follows:
`prefix qudt: <http://qudt.org/2.1/schema/qudt/> `
However, VOCAB_QUDT-CITATION-v2.1/ttl defines that prefix as follows:
`prefix qudt: <http://qudt.org/2.1/schema/qudt#> `

In the same file, I also found this snippet that uses the "#" separator:
```
citation:PubKind
  a owl:Class ;
  rdfs:label "Pub kind" ;
  rdfs:subClassOf <http://qudt.org/schema/qudt/StructuredDatatype> ;
```

The file VOCAB_QUDT-NIDR-v2.1.ttl contains the following line:
` vaem:namespace "http://qudt.org/schema/qudt#" ;`
that, again, using the "#" separator instead of the more common "/".

Are there issues if we uniformly use "/" instead of "#" as separator? Many thanks.

steveraysteveray:
The two vocabularies you mention are files that we haven't integrated into our current [release](https://github.com/qudt/qudt-public-repo/releases) of QUDT. The # should be a /. There are also a few other problems in those files - we'll get to them eventually! Thanks for pointing this out.

PR #685, opened 2023-04-18
Fix namespace IRI of citation:

ektrah:
As discussed in https://github.com/qudt/qudt-public-repo/pull/681, `<http://qudt.org/2.1/schema/citation#>` should be the versionless namespace IRI `<http://qudt.org/schema/citation/>`.

`<http://qudt.org/2.1/schema/citation#PubKinds>` and `<http://qudt.org/2.1/schema/citation#PubKind>` are not referenced anywhere and don't seem to have a clear purpose, so maybe they could just be removed?

### DIFF
--- a/schema/SCHEMA_QUDT-CITATION-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-CITATION-v2.1.ttl
@@ -215,18 +215,6 @@ citation:Citation
     ] ;
   skos:definition "Used to define bibliographic references, compatible with BibTeX and BibLaTeX characteristics" ;
 .
-citation:PubKind
-  a owl:Class ;
-  rdfs:label "Pub kind" ;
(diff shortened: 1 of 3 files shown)


ektrah:
Related to https://github.com/qudt/qudt-public-repo/issues/481

steveraysteveray (commented):
We're going to have to discuss this in the Board. I agree with removing the version number from the prefix definition, but the other removals might be premature. Earlier, we had intended to use the Citation definitions to provide more structure to a citation. There were some problems in the implementation, so we temporarily stopped using them. However, we may revisit this in which case the definitions would be needed (again).

ektrah:
I only came across `SCHEMA_QUDT-CITATION-v2.1.ttl` because I was spending time on `SCHEMA_QUDT-DATATYPE-v2.1.ttl` 🙂

Please let me know if I should revert one of the two commits and which one.

jhodgesatmb (approved):

PR #689, opened 2023-04-19
Fix the ontology IRI of QUDT Sources Vocabulary

ektrah:
I didn't try to fix the occurrences in `vocab\VOCAB_QUDT-SOURCES-v2.0.ttl` and `vocab\quantitykinds\to-be-deprecated-quantitykinds\VOCAB_QUDT-QUANTITY-KINDS-NIST-CONSTANTS-v2.0.ttl`.

### DIFF
--- a/vocab/VOCAB_QUDT-SOURCES-v2.1.ttl
+++ b/vocab/VOCAB_QUDT-SOURCES-v2.1.ttl
@@ -70,19 +70,19 @@ CODATA sponsors the CODATA international conference every two years. [Wikipedia]
   qudt:name "CODATA" ;
   qudt:url "http://en.wikipedia.org/wiki/Committee_on_Data_for_Science_and_Technology"^^xsd:anyURI ;
   qudt:url "http://www.sizes.com/units/codata.htm"^^xsd:anyURI ;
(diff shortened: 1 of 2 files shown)


ektrah:
Oh, I see. This PR was prompted by the reference in `VOCAB_QUDT-CONSTANTS-v2.1.ttl` to `<http://qudt.org/2.0/vocab/sources>`. I assumed that this term never shows up as a subject because the statements in `VOCAB_QUDT-SOURCES-v2.1.ttl` are about `<http://qudt.org/2.1/vocab/sources>`. Now I understand that changing `<http://qudt.org/2.0/vocab/sources>` to `<http://qudt.org/2.1/vocab/sources>` in `VOCAB_QUDT-CONSTANTS-v2.1.ttl` won't fix that, since `VOCAB_QUDT-SOURCES-v2.1.ttl` is actually not part of the release.

Taking a closer look, the statements about `qudt:PhysicsForums` in `VOCAB_QUDT-CONSTANTS-v2.1.ttl` seem to be duplicated from `VOCAB_QUDT-SOURCES-v2.1.ttl`. Maybe they could just be removed?

PR #693, opened 2023-04-26
request to add new unit:MegaGM-PER-HA

elisagirola:


dr-shorthair:
Already there in practice - http://qudt.org/vocab/unit/TONNE-PER-HA

steveraysteveray (changes requested):
The dimension vector should be qkdv:A0E0L-2I0M1H0T0D0  (i.e. per unit area, not per unit volume).
The quantity kind should be quantitykind:MassPerArea

Also, please add the following two lines:
  qudt:exactMatch unit:TON_Metric-PER-HA ;
  qudt:exactMatch unit:TONNE-PER-HA ;

The unece Code you list is for Mega gram per cubic metre, so that isn't right. I couldn't find a suitable unece Code, so you should just delete that line.

steveraysteveray (changes requested):
Sorry to drag this out, but I forgot to check the iec61360 code, and it seems that one is for Mega gram per metre cubed, so I think we need to delete that line as well.

steveraysteveray (approved):

jhodgesatmb:
Instead of deleting unece and iec61360 codes because they are wrong, we should strive to find out the appropriate codes. Deleting them takes this unit out of sight and thus out of mind.

steveraysteveray:
I strived (strove?) and failed to find them.

elisagirola:
Thanks All. Much appreciated

Issue #704, opened 2023-05-17
Undefined terms in QUDT release


ektrah:
QUDT currently uses a number of terms as predicates and/or objects, that however never show up as subjects. This is often (but not always) an indication that there is typo or wrong prefix in a Turtle file. Here is the list:

| Undefined Term | |
|:-|:-|
| `http://qudt.org/2.1/vocab/dimension`                                  |                                                                        |
| `http://qudt.org/2.1/vocab/sources`                                    |                                                                        |
| `http://qudt.org/schema/qudt/AtomicUnit`                               |                                                                        |
| `http://qudt.org/schema/qudt/baseCGSUnitDimensions`                    |                                                                        |
| `http://qudt.org/schema/qudt/baseImperialUnitDimensions`               |                                                                        |
| `http://qudt.org/schema/qudt/baseISOUnitDimensions`                    |                                                                        |
| `http://qudt.org/schema/qudt/baseSIUnitDimensions`                     |                                                                        |
| `http://qudt.org/schema/qudt/baseUSCustomaryUnitDimensions`            |                                                                        |
| `http://qudt.org/schema/qudt/BooleanStateType`                         |                                                                        |
| `http://qudt.org/schema/qudt/DataElement`                              |                                                                        |
| `http://qudt.org/schema/qudt/DatatypeStructuredData`                   |                                                                        |
| `http://qudt.org/schema/qudt/descriptor`                               |                                                                        |
| `http://qudt.org/schema/qudt/DIM_CGS-EMU_L`                            |                                                                        |
| `http://qudt.org/schema/qudt/DIM_CGS-ESU_L`                            |                                                                        |
| `http://qudt.org/schema/qudt/DIM_CGS-L`                                |                                                                        |
| `http://qudt.org/schema/qudt/DIM_CGS-L2`                               |                                                                        |
| `http://qudt.org/schema/qudt/DIM_ISO-L-3`                              |                                                                        |
| `http://qudt.org/schema/qudt/DIM_ISO-L2MT-2`                           |                                                                        |
| `http://qudt.org/schema/qudt/DIM_ISO-MT-2`                             |                                                                        |
| `http://qudt.org/schema/qudt/DIM_Planck-L`                             |                                                                        |
| `http://qudt.org/schema/qudt/DIM_SI-L`                                 |                                                                        |
| `http://qudt.org/schema/qudt/DIM_SI-m-3`                               |                                                                        |
[...]

PR #705, opened 2023-05-18
Fix some undefined terms

ektrah:
Here's my attempt at resolving a first portion of #704. It might be easier to review the individual commits than the whole pull request. Please let me know if I'm on the right track :blush:

### DIFF
--- a/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
@@ -143,9 +143,9 @@ qudt:AlignmentType
   rdfs:label "Alignment type" ;
   rdfs:subClassOf qudt:EnumeratedValue ;
   owl:oneOf (
-      qudt.type:BitAligned
-      qudt.type:ByteAligned
-      qudt.type:WordAligned
+      qudt:BitAligned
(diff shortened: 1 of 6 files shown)


steveraysteveray:
These changes all seem reasonable. I will wait until my QUDT colleagues have a chance to weigh in though.

Also, be aware that we are currently discussing the modeling of datatypes in SHACL, which raises a number of issues, so things may be changing. But (in my opinion), it's good to clean house with what we have anyway.

steveraysteveray:
We finally got a chance to discuss this and we're fine with your suggestions except for changing the prefix qudt.type to qudt. This is because we generally use qudt: as a prefix for schema entities (classes, properties...) and other prefixes for the vocabulary/instances (e.g. unit:, quantitykind:, constant:). After some back and forth, we settled on qudt_type: as the preferred prefix name rather than qudt.type or just qudt. If you make that change, I think we're good.

fkleedorfer:
@steveraysteveray how about we wrap this up (merge, resolve conflicts, make the suggested change (qudt.type -> qudt) and be done with it?

fkleedorfer:
I have to think about this more...

ektrah:
Replaced by  #1248

PR #710, opened 2023-06-15
fixed small typo in unit:M3-PER-HA qudt:symbol

kronmar:
Found a small inconsistency in the unit:M3-PER-HA. It's the only one where the object associated with predicate qudt:symbol uses latex-style "^" for superscript instead of the actual superscript.


Rdataflow:
would be great to have this in for consistent UX presentation among other units 👍

steveraysteveray (approved):
Thanks for catching this.

PR #737, opened 2023-08-11
Update prefixes to Dec. 2022 BIPM SI 9th Ed. V2.01

DocMT:
New prefixes have been added in the December 2022 update of the BIPM SI Brochure (9th Edition, December 2022, V2.01), specifically (see https://www.bipm.org/en/measurement-units/si-prefixes):

1.) quetta (symbol: "Q"; multiplying factor 10^30),
2.) ronna (symbol: "R"; multiplying factor 10^27),
3.) ronto (symbol: "r"; multiplying factor 10^-27) and
4.) quecto (symbol: "q"; multiplying factor 10^-30)

Note: I do not know if the updates to the GMD_QUDT-PREFIXES metadata and perfixes version bump are acceptable, please check.

### DIFF
--- a/vocab/prefixes/VOCAB_QUDT-PREFIXES-v2.1.ttl
+++ b/vocab/prefixes/VOCAB_QUDT-PREFIXES-v2.1.ttl
@@ -13,7 +13,7 @@
 <http://qudt.org/2.1/vocab/prefix>
   a owl:Ontology ;
   vaem:hasGraphMetadata vaem:GMD_QUDT-PREFIXES ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (changes requested):
Thanks for the updates!

Actually, could you put the 2.1.30 occurrences back to 2.1.29? They get automatically updated with each Release and the script would get messed up otherwise.

DocMT:
Thank you for your quick review and for the great project! Please let me know if anything else needs to be changed.

PR #739, opened 2023-08-14
Update descriptions

connorjcantrell:
Just pushed a quick update to the descriptions for both the US Ton and the Imperial Gallon. I attempted to give them a little more clarity and context.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -7480,7 +7480,7 @@ unit:GALILEO
 .
 unit:GAL_IMP
   a qudt:Unit ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (changes requested):
A couple of comments. For the GAL_IMP I would say "approximately 1.2 U.S. Gallons." 

For the TON_US, how about "A ton is a unit of mass in the US customary system, where 1 ton is equal to 2000 pounds of mass."  Note that the dimensionality is Mass (A0E0L0I0M1H0T0D0), not force (A0E0L1I0M1H0T-2D0). Weight has to do with gravitational attraction, rather than inertial mass. While it's true that here on earth, people often conflate the two, we strive to be precise in QUDT!

There is a unit of force called unit:TON_F_US.

steveraysteveray (approved):

PR #747, opened 2023-09-02
Adding Catalytic Activity Concentration Units

glow-mdsol:
I've added these in a separate file for review.  Catalytic Activity Concentrations are used in Lab results.  I've added the new units (based on the UCUM codes) and added a reference to IU/L (which is another unit used to represent the results)

### DIFF
--- a/submissions/catalytic_activity_concentrations.ttl
+++ b/submissions/catalytic_activity_concentrations.ttl
@@ -1,5 +1,6 @@
 @prefix quantitykind: <http://qudt.org/vocab/quantitykind/> .
 @prefix qudt: <http://qudt.org/schema/qudt/> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 
 
 
--- a/submissions/catalytic_activity_concentrations.ttl
+++ b/submissions/catalytic_activity_concentrations.ttl
@@ -21,7 +21,7 @@ quantitykind:CatalyticActivityConcentration
   skos:broader <http://qudt.org/vocab/quantitykind/AmountOfSubstanceConcentrationOfB> ;
 .
 
-# This amends the existing unit to link to the new QK
+# This amends the existing unit to link to the QK
 unit:IU-PER-L
   a qudt:DerivedUnit ;
   qudt:hasQuantityKind quantitykind:CatalyticActivityConcentration .
@@ -93,7 +93,7 @@ unit:NanoKAT-PER-L
   qudt:symbol "nkat/L" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Nanokatal Per Litre"@en ;
-  rdfs:label "Nanokatal Per Litre"@en-us
+  rdfs:label "Nanokatal Per Litre"@en-us ;
 .
 
 unit:PicoKAT-PER-L
--- a/submissions/catalytic_activity_concentrations.ttl
+++ b/submissions/catalytic_activity_concentrations.ttl
@@ -52,7 +52,7 @@ unit:MilliKAT-PER-L
   dcterms:description "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
   qudt:plainTextDescription "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
   qudt:applicableSystem sou:SI ;
-  qudt:conversionMultiplier 1.0  ;
+  qudt:conversionMultiplier 0.000001  ;
   qudt:definedUnitOfSystem sou:SI ;
   qudt:derivedCoherentUnitOfSystem sou:SI ;
   qudt:hasQuantityKind quantitykind:CatalyticActivityConcentration ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (changes requested):
Are you sure about the conversion multipliers? I ask because the SI unit of volume is the cubic metre, not the litre. So that would imply unit:KAT-PER-L would have qudt:conversionMultiplier of 0.001, to convert to KAT-PER-M3. Is that right?

You also make reference to http://qudt.org/vocab/quantitykind/AmountOfSubstanceConcentration, but that is not present in our quantity kind vocabulary. Perhaps you can help me understand how this differs from http://qudt.org/vocab/quantitykind/AmountOfSubstanceConcentrationOfB, which we do have. This is not my field of expertise!

Once we're settled on all this, I'm happy to cut and paste your submissions into the main vocabulary files.

Thanks!

glow-mdsol:
That QuantityKind is weird - I had what you showed and then changed it.  Also identified a couple of copy-paste errors.  I think the conversions are now aligned with expectations

submissions/catalytic_activity_concentrations.ttl
```diff
@@ -0,0 +1,113 @@
...
+  rdfs:label "Katal Per Litre"@en-us 
+.
+
+unit:MilliKAT-PER-L
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  dcterms:description "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
+  qudt:plainTextDescription "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
+  qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 1.0  ;
```
steveraysteveray, line 55: 1 MilliKAT-PER-L is 0.000001 KAT-PER-M3.  It's 1000 times smaller than a KAT-PER-L, not 1000 times bigger.

submissions/catalytic_activity_concentrations.ttl
```diff
@@ -0,0 +1,113 @@
...
+  rdfs:label "Millikatal Per Litre"@en-us ;
+.
+
+unit:MicroKAT-PER-L
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  dcterms:description "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
+  qudt:plainTextDescription "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
+  qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 1000.0  ;
```
steveraysteveray, line 72: Should be 0.000000001

submissions/catalytic_activity_concentrations.ttl
```diff
@@ -0,0 +1,113 @@
...
+  rdfs:label "Microkatal Per Litre"@en ;
+  rdfs:label "Microkatal Per Litre"@en-us ;
+.
+
+unit:NanoKAT-PER-L
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  dcterms:description "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
+  qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 1000000.0 ;
```
steveraysteveray, line 88: 0.000000000001

submissions/catalytic_activity_concentrations.ttl
```diff
@@ -0,0 +1,113 @@
...
+  rdfs:label "Nanokatal Per Litre"@en ;
+  rdfs:label "Nanokatal Per Litre"@en-us ;
+.
+
+unit:PicoKAT-PER-L
+  a qudt:DerivedUnit ;
+  a qudt:Unit ;
+  dcterms:description "A unit of catalytic activity used especially in the chemistry of enzymes. A catalyst is a substance that starts or speeds a chemical reaction. Enzymes are proteins that act as catalysts within the bodies of living plants and animals. A catalyst has an activity of one katal if it enables a reaction to proceed at the rate of one mole per second. "^^rdf:HTML ;
+  qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 1000000000.0 ;
```
steveraysteveray, line 104: 0.000000000000001

steveraysteveray (changes requested):

submissions/catalytic_activity_concentrations.ttl
glow-mdsol: Updated

submissions/catalytic_activity_concentrations.ttl
glow-mdsol: Updated

submissions/catalytic_activity_concentrations.ttl
glow-mdsol: Updated

submissions/catalytic_activity_concentrations.ttl
glow-mdsol: Updated

PR #748, opened 2023-09-08
Added UCUM code for FRACTION

glow-mdsol:
* Added missing UCUM code for FRACTION
* Added URI for om ontology


steveraysteveray (changes requested):
unit:FRACTION does not correspond directly to http://www.ontology-of-units-of-measure.org/resource/om-2/AmountOfSubstanceFractionUnit, because it is a more general unit. OM has a bunch of specializations of FRACTION, but no general unit for FRACTION. For this reason, I think we should leave out the omUnit reference.

glow-mdsol:
Understood, updated

steveraysteveray (approved):

PR #749, opened 2023-09-11
Retiring owl:sameAs

steveraysteveray:
...as it causes problems with OWL reasoners.


dr-shorthair:
In some cases they may be replaced by 
```
owl:equivalentClass
owl:equivalentProperty
```

steveraysteveray:
True. For now, we have qudt:exactMatch instead. In the future, if a need for one of the owl or skos relations becomes clear, a quick SPARQL query could fix that.

jhodgesatmb:
Has the use of qudt:exactMatch eliminated the problem? So the new policy is that only qudt:exactMatch will be used and you have already proliferated this change throughout the schema and vocabularies? And which OWL reasoners were affected?

steveraysteveray:
1. Will know whether this has eliminated the problem with the next Release, because the graphs are retrieved from the web.
2. Yes, the change has been made on all files
3. HermiT 1.4.3.456

ralphtq (approved):

PR #758, opened 2023-09-22
Add 6 units and one quantity kind

mgberg:
This adds `quantitykind:SpecificPower` and declares existing relevant units to have that quantity kind.

Furthermore, the following units are added: `unit:IN-PER-MIN`, `unit:J-PER-GM-K`, `unit:MegaPSI`, `unit:MilliJ-PER-GM`, `unit:MilliW-PER-MilliGM` and `unit:W-PER-GM`.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -19673,7 +19673,7 @@ unit:MilliJ-PER-GM
   qudt:applicableSystem sou:CGS-EMU ;
   qudt:applicableSystem sou:CGS-GAUSS ;
   qudt:applicableSystem sou:SI ;
-  qudt:conversionMultiplier 1000000000.0 ;
+  qudt:conversionMultiplier 1.0 ;
   qudt:hasDimensionVector qkdv:A0E0L2I0M0H0T-2D0 ;
   qudt:hasQuantityKind quantitykind:SpecificEnergy ;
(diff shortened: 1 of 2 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -19607,6 +19666,23 @@ unit:MilliJ
...
   rdfs:label "Millijoule"@en ;
 .
+unit:MilliJ-PER-GM
+  a qudt:Unit ;
+  dcterms:description "The 0.001-fold of the SI base unit joule divided by the 0.001-fold of the SI base unit kilogram."^^rdf:HTML ;
+  qudt:applicableSystem sou:CGS ;
+  qudt:applicableSystem sou:CGS-EMU ;
+  qudt:applicableSystem sou:CGS-GAUSS ;
+  qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 1000000000.0 ;
```
steveraysteveray, line 19676: Multiply top and bottom by 1000 and you get J-PER-KiloGM, so the conversionMultiplier must be 1.0

steveraysteveray (changes requested):

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
mgberg: How did I manage to do that? 🤦‍♂️ Fixed.

steveraysteveray (approved):

PR #772, opened 2023-10-04
Add new units from existing quantity kinds and extend various ucum symbols

jimkont:
Added the following units, which were mostly copied from existing related units:
* unit:IU-PER-MilliL
* unit:KAT-PER-MicroL
* unit:MicroGM-PER-deciL
* unit:MicroGM-PER-MilliL
* unit:NanoGM-PER-DeciL
* unit:GM-PER-DeciL
* unit:NUM-PER-MilliM3
* unit:NUM-PER-NanoL
* unit:NUM-PER-PicoL

in addition, added some missing or alternative ucum symbols in other units.

One question I have is related to the units unit:IU-PER-MilliL (new) and unit:IU-PER-L (existing). in the existing one the description says `The magnitude of one <strong>IU/L</strong> depends on the material, so there is no single conversion multiplier. (See <a href=\"http://qudt.org/vocab/unit/IU)\"><strong>IU</strong></a> for more information)`. 
To me it seems that the international unit `IU` is generic and the `quantitykind:AmountOfSubstancePerUnitVolume` could be assigned to both of them (and thus, define the related multipliers) but I wanted to verify this assumption first.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -17832,7 +17832,7 @@ unit:MicroGM-PER-L
   rdfs:label "Microgram Per Liter"@en-us ;
   rdfs:label "Microgram Per Litre"@en ;
 .
-unit:MicroGM-PER-deciL
+unit:MicroGM-PER-DeciL
   a qudt:Unit ;
   qudt:applicableSystem sou:CGS ;
   qudt:applicableSystem sou:CGS-EMU ;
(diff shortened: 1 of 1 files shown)


jimkont:
Is there any additional information I should provide in this PR, or should I maybe split it into separate ones per quantity kind?

steveraysteveray:
No need to split it. Perhaps you hadn't noticed that I had requested one minor change. It is much easier for you to make that change because I don't have access to update your branch.

jimkont:
thanks! unfortunately, I cannot find your comment, could you please point it to me, or request a new change?

steveraysteveray:
It is

unit:MicroGM-PER-deciL should be unit:MicroGM-PER-DeciL

jimkont:
Done! regarding my question on adding the `quantitykind:AmountOfSubstancePerUnitVolume` in `unit:IU-PER-L` and the new `unit:IU-PER-MilliLL` together with a conversion multiplier, do you have any feedback?
I noticed that `unit:IU-PER-MilliGM` which is related has `quantitykind:AmountOfSubstancePerUnitMass` assigned and also a conversion multiplier, and that is why I ask

steveraysteveray:
Good catch! After researching IU and documenting unit:IU, I was not comprehensive in updating all the instances involving IU. None of those instances should have conversion multipliers. You are invited to make those changes in this PR, or we could go ahead and approve this PR and either you or I could make the other fixes later.

What's your pleasure?

PR #777, opened 2023-10-09
Reflectivity and Reflectance

mgberg:
This includes the following changes:
- Add `Reflectivity` as a narrower Quantity Kind than `Reflectance`.
- Add 3 units for `Reflectance` (and therefore `Reflectivity`).
- Remove a lingering reference to `quantitykind:Refectance`.
- Add a conversion multiplier of 0.01 to `unit:PERCENT`.

### DIFF
--- a/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
+++ b/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
@@ -17605,16 +17605,17 @@ quantitykind:RecombinationCoefficient
 .
 quantitykind:Reflectance
   a qudt:QuantityKind ;
-  dcterms:description "Reflectance and reflectivity generally refer to the fraction of incident power that is reflected at an interface, while the term \"reflection coefficient\" is used for the fraction of electric field reflected. Reflectance is always a real number between zero and 1.0."^^rdf:HTML ;
+  dcterms:description "Reflectance generally refers to the fraction of incident power that is reflected at an interface, while the term \"reflection coefficient\" is used for the fraction of electric field reflected. Reflectance is always a real number between zero and 1.0."^^rdf:HTML ;
   qudt:applicableUnit unit:UNITLESS ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
   qudt:informativeReference "http://en.wikipedia.org/wiki/Reflectivity"^^xsd:anyURI ;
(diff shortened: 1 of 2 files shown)


steveraysteveray:
@mgberg, I like the thorough descriptions and distinctions between Reflectivity and Reflectance. I do have a question concerning the applicableUnits. The way you have the skos:broader relations defined, Reflectance will have all the dimensionless ratio applicable units, even though your description states reflectance is always between 0.0 and 1.0. Is that OK with you, as a domain expert? The alternative is to have the suitable units just refer directly to Reflectance and to Reflectivity as appropriate. 

If so, 
Would unit:FRACTION be the best one for Reflectance?
Do you want to have explicit real and imaginary scalar units for Reflectivity? (We do that for electrical AC power: ActivePower and ReactivePower)

Eventually we plan to ramp up our use of datatypes beyond scalars...

What say you?

mgberg:
@steveraysteveray Good question. I guess it partially depends on how the different units are intended to be used. We have data in which some sources report "fractional" values (including for Reflectivity/Reflectance) as a unitless number (i.e. 0-1) and other sources report the same kind of values as a percent (i.e. 0-100). Consequently, I have internally defined that `unit:PERCENT` has a conversion multiplier of `0.01` such that we can convert between `unit:PERCENT` and `unit:FRACTION` or `unit:UNITLESS` as needed. I'd actually be curious to know what you think about adding that conversion multiplier (unless we're using it wrong 😆).

Also, are some units intended to only have values between 0 and 1 (or 0 and 100, or some other range)? I don't seem to recall that being captured or described anywhere but I could be wrong. It could be a good capability.

As for including all the units- as I mentioned above, we have lots of different sources reporting values in different units depending on the context (e.g. length scale), and dimensionless numbers are one of the worst offenders (which is why a while ago I had to add `unit:PPTM`). For that reason, I tend to lean toward being more flexible than less. However, if the community disagrees with that premise, I'd be happy to apply a restricted set of units for Reflectance including at least `UNITLESS` and `PERCENT`.

As described [here](https://en.wikipedia.org/wiki/Fresnel_equations#Power_(intensity)_reflection_and_transmission_coefficients), and [here](https://en.wikipedia.org/wiki/Reflectance#Reflectivity), reflectivity/reflectance should always be a positive real number.

steveraysteveray:
Responding from the bottom up:

I'm leaning toward restricting the units for Reflectance, mostly because of the disconnect between the description and the applicableUnit values. Let's see if others have opinions. If we go that way, then you would need distinct references from Reflectivity to quantity kinds.

Regarding restricting values (0-1 or 0-100), we do have qudt:floatPercentage as a datatype. It's defined that way in the OWL schema, but the OWL stuff crept into the SHACL version as well, which needs fixing. We could do something similar for 0-1.

Finally, I support your suggestion of the conversion multiplier for unit:PERCENT to be 0.01. Makes perfect sense to me.

mgberg:
Yup, I missed `qudt:floatPercentage`- that might be useful.

I have no problem with being more restrictive- I've updated this to include 3 units explicitly.

I also added the conversion multiplier to `PERCENT`.

steveraysteveray (approved):

PR #797, opened 2023-10-27
Standardize qudt:symbol for a few units

mgberg:
Just standardizes a few unit symbols to match the rest of the symbols.

Note that I used `TOR` for Ton of Refrigeration because it seems to be commonly used but doesn't conflict with the symbol for Register Ton.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -11168,7 +11168,7 @@ unit:Kilo-FT3
   qudt:hasDimensionVector qkdv:A0E0L3I0M0H0T0D0 ;
   qudt:hasQuantityKind quantitykind:LiquidVolume ;
   qudt:hasQuantityKind quantitykind:Volume ;
-  qudt:symbol "kft³" ;
+  qudt:symbol "k(ft³)" ;
   qudt:ucumCode "[k.cft_i]"^^qudt:UCUMcs ;
   qudt:ucumCode "k[ft_i]3"^^qudt:UCUMcs ;
   qudt:uneceCommonCode "FC" ;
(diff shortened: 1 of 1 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -11101,7 +11101,7 @@ unit:Kilo-FT3
   qudt:hasDimensionVector qkdv:A0E0L3I0M0H0T0D0 ;
   qudt:hasQuantityKind quantitykind:LiquidVolume ;
   qudt:hasQuantityKind quantitykind:Volume ;
-  qudt:symbol "k.ft³" ;
+  qudt:symbol "kft³" ;
```
steveraysteveray, line 11104: What would be the symbol for (1000 feet)**3, as compared with this unit, which is 1000 (cubic feet)?

steveraysteveray: Either variant of the symbol still leaves me with the question of how to interpret it unambiguously. It seems to me that the only recourse is the use of parentheses, if we are to include a symbol at all. What about k(ft³)?

steveraysteveray (changes requested):
Something strange happened, and the previous xsd:double versions of conversionMultiplier for some units (e.g. PlanckPressure and other Planck units) have reappeared. Everything else seems fine. Could you fix?

mgberg:
> Something strange happened, and the previous xsd:double versions of conversionMultiplier for some units (e.g. PlanckPressure and other Planck units) have reappeared. Everything else seems fine. Could you fix?

@steveraysteveray That's strange! When I open the file in Composer, it converts from the scientific notation to the long format. Is there a setting in Composer I'm missing to prevent that from happening? If not, that'll make it challenging to use Composer going forward to edit units.

steveraysteveray (approved):

steveraysteveray:
@mgberg, I could not recreate your problem, but I see you fixed it. What was the issue?

PR #803, opened 2023-11-03
Unit and Quantity Kind Label Language Tags

mgberg:
A small fix to ensure all labels for units and quantity kinds have a language tag.

Note that one commit in this PR is just reserializing the quantity kinds file since it did not match Composer's normal output but contained no modifications otherwise.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -3307,7 +3307,7 @@ unit:CD-PER-LM
   qudt:hasQuantityKind quantitykind:LuminousIntensityDistribution ;
   qudt:symbol "cd/lm" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
-  rdfs:label "Candela per Lumen" ;
+  rdfs:label "Candela per Lumen"@en ;
 .
 unit:CD-PER-M2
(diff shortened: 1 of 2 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -16562,7 +16562,8 @@ unit:MOL-PER-M2-DAY
   qudt:symbol "mol/(m²⋅day)" ;
   qudt:ucumCode "mol.m-2.d-1"^^qudt:UCUMcs ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
-  rdfs:label "Moles per square metre per day" ;
+  rdfs:label "Moles per Square Metre per Day"@en ;
+  rdfs:label "Moles per Square Metre per Day"@en-us ;
```
steveraysteveray, line 16566: Change "Metre" to "Meter"

steveraysteveray (changes requested):
One tiny change requested. See embedded comment.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
mgberg: That was my intent 🤦‍♂️ I'll fix that now

steveraysteveray (approved):

PR #804, opened 2023-11-06
feat: add the unit decisecond

fennibay:
This unit is generally used by BACnet protocol, so it's generally relevant for building automation applications. It could also be useful elsewhere, and for the sake of completeness.


fennibay:
Thanks for merging so fast @steveraysteveray

PR #816, opened 2023-11-17
Edi contributions, 38 units

mobb:
Here is our first contribution, 38 units used in the environmental sciences, which can be matched to datasets from the https://edirepository.org and NEON (https://www.neonscience.org/). 
We suspect that several units will not validate. We are using the Protege editor (5.6.3), and so far, have not gotten the installed reasoners (Pellet, HermiT) to display errors, such as a unit whose dimension vector does not match the quantityKind assigned.  Here is one suspicious unit: NUM-PER-KiloGM. 
We also think our application of quantityKinds is inconsistent, and we are still looking for a good way to visualize the structure, including and relationships so we can fine tune this.


steveraysteveray:
I have been swamped with other tasks lately, but one quick observation. I see you have marked 8 units with "unknown" quantity kinds. I'm thinking that if indeed it is not possible to know the quantity kind without the context of the measurement, that is what the concept of qudt:Quantity is all about. The qudt:QuantityKind is more of a context-free characterization of the kind of measurement, so you should choose or invent one like that. See the wiki introduction [here](https://github.com/qudt/qudt-public-repo/wiki/User-Guide-for-QUDT).

steveraysteveray:
We can go over these in more detail if you have questions. Sometimes the problems are simple errors like TitleCase instead of lowercase URIs. Sometimes they are more subtle. 

For example, at present, the value of qudt:plainTextDescription must be a simple string without any language tag. (We might relax that and allow multiple languages as we do with rdfs:label.)

Another place you have two entries for the same unit that causes some cardinality violations when they are combined. 

Another case: some quantitykinds instances have qudt:exactMatch relations with others. In those cases, any units referring to one quantitykind must also refer to the other as well (Velocity = LinearVelocity, so IN-PER-YR must refer to both)

See if the following gets you started.

image
image

Steve

steveraysteveray (changes requested):
See my comments [here](https://github.com/qudt/qudt-public-repo/pull/816#issuecomment-1845475900)

mobb:
no longer needed.

PR #829, opened 2023-12-05
 Fixed invalid HTML flagged by Gabe Fierro

steveraysteveray:


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -28572,7 +28572,7 @@ unit:TON_Assay
 .
 unit:TON_FG
   a qudt:Unit ;
-  dcterms:description "12000 btu per hour</p>"^^rdf:HTML ;
+  dcterms:description "12000 btu per hour"^^rdf:HTML ;
```
ralphtq, line 28575: As is not need to be an rdf:HTML. The change could have been done by adding <p> at the start of the text.

ralphtq (changes requested):

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
steveraysteveray: True, but many of our dcterms:description values do not contain any HTML markup, but are typed as rdf:HTML. Currently, schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl specifies a range of rdf:HTML. We could remove that constraint, in which case the values could be simple strings or HTML.

PR #832, opened 2023-12-07
Fix qudt:symbol entries

fkleedorfer:
Assuming that base unit symbols are correct, unless it is obvious, symbols are generated according to these rules and replace existing symbols, except in some cases in which there seems to be an intentional use of a very different symbol (such as 'mil' or 'mas').

1. use the symbol of the base unit
2. multiply with a middle dot
3. divide with a slash
4. enclose multiplication in the denominator with parentheses
5. use the same order of units as is used in the localname


steveraysteveray (approved):
Wow, you have caught and fixed so many errors in the symbols that I'm going to approve this, even if a few new errors might have slipped in. Thanks!

fkleedorfer:
We could think about generating symbols, ucum codes, and latex symbols automatically when they are missing. Thus, nonstandard symbols can be realized while changes to the base unit symbols propagate to derived units

PR #833, opened 2023-12-07
Fix qudt:hasDimensionVector entries

fkleedorfer:
For each fix, a comment is provided that contains the 'factor unit tree' along with the dimension vectors of each the factor, so that reviewers can check the combination of those vectors and verify the error in the existing DV. These comments should be removed after verification of the replaced DVs.

All other dimension vectors are correct or cannot be verified because their factor units' dimension vectors are missing.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -224,7 +224,17 @@ unit:A-PER-J
   qudt:conversionMultiplier 1.0 ;
   qudt:derivedCoherentUnitOfSystem sou:SI ;
   qudt:expression "\\(A/J\\)"^^qudt:LatexString ;
-  qudt:hasDimensionVector qkdv:A0E1L-2I0M-1H0T3D0 ;
+  # factor unit tree for unit A/J
+  #     +-- unit:A-PER-J                  	http://qudt.org/vocab/dimensionvector/A0E1L-2I0M-1H0T3D0
+  #     |   +-- unit:J^-1                     	http://qudt.org/vocab/dimensionvector/A0E0L-2I0M-1H0T2D0
+  #     |   |   +-- unit:N                        	http://qudt.org/vocab/dimensionvector/A0E0L1I0M1H0T-2D0
+  #     |   |   |   +-- unit:KiloGM                   	http://qudt.org/vocab/dimensionvector/A0E0L0I0M1H0T0D0
+  #     |   |   |   +-- unit:M                        	http://qudt.org/vocab/dimensionvector/A0E0L1I0M0H0T0D0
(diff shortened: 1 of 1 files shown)


steveraysteveray (approved):
You seem to have a very handy algorithm working here. I had been doing this manually, but clearly this is much more comprehensive. I note that your algorithm seems to transform "energy" into Newton-Metres. Personally, I use e=mc**2, which translates to base units directly, but the computer doesn't really care! Any chance you could share your code, or even somehow turn it into a SHACL validation constraint?

steveraysteveray:
I now realize that we have a cascading set of errors. You have fixed the unit's dimension vector, but now of course, the associated quantity kind either

1. Is the correct quantity kind, but it also has the wrong dimension vector, or
2. Is the wrong quantity kind.

Would you be willing to chase these down? They show up in the validation check.

steveraysteveray:
image
image

fkleedorfer:
The algorithm is simple once you have the factor unit hierarchy: take the dim vector to the power of the unit, combine vectors in the parent node, and continue that recursively.

I am doing this in java, it is probably doable with shacl sparql functions and the preprocessing I am doing with a couple of construct queries and a few extra triples. However, I don't have the time to do that now. Feel free to use qudtlib to do it though ( release with that code in the next few days)

fkleedorfer:
The algorithm is simple once you have the factor unit hierarchy: take the dim vector to the power of the unit, combine vectors in the parent node, and continue that recursively.

I am doing this in java, it is probably doable with shacl sparql functions and the preprocessing I am doing with a couple of construct queries and a few extra triples. However, I don't have the time to do that now. Feel free to use qudtlib to do it though ( release with that code in the next few days)

steveraysteveray:
There seems to be a problem with your factor logic. For example, you had:
```
  # factor unit tree for unit µg/cm²
  #     +-- unit:MicroG-PER-CentiM2       	http://qudt.org/vocab/dimensionvector/A0E0L-2I0M1H0T0D0
  #     |   +-- unit:MicroG                   	http://qudt.org/vocab/dimensionvector/A0E0L1I0M0H0T-2D0
  #     |   +-- unit:CentiM^-2                	http://qudt.org/vocab/dimensionvector/A0E0L-2I0M0H0T0D0
  #
  qudt:hasDimensionVector qkdv:A0E0L-1I0M0H0T-2D0 ;
```

steveraysteveray:
Aha, I now see that we had an error in the URI: MicroG-PER-CM2 should have been MicroGM-PER-CM2. Your algorithm took it at face value and assumed we were talking about MicroGravity per cm**2, and thus computed the appropriate DV.

So we should fix that first, and then tackle the cascading issues. It may be that the qk algorithm may need to involve a human to make the judgements... I can't go any further right now - maybe tomorrow.

fkleedorfer:
I've added some information about the quantitykinds - does not seem to be  an easy decision in each case

PR: #836

fkleedorfer:
No, actually: #837

PR #834, opened 2023-12-07
Deprecate unit:ERG-PER-G

fkleedorfer:
The correct unit is unit:ERG-PER-GM.


steveraysteveray:
@fkleedorfer, I have updated the vocabulary files so that they pass all the validation tests in #838, which has been pushed to the main branch. So now we have a good new starting point. Would you mind merging main into this branch and resolving the merge conflicts? Then I will be able to evaluate this and the other PRs more easily.

Thanks.

steveraysteveray (approved):
Several of your unit definitions are duplicates of existing entries. This is not inherently a problem, since they will be combined when re-serialized by tools like TopBraid Composer. My impression is that you are using a simple editor, which is harder to do this with. So, I'm going to go ahead and approve/merge, remove the duplicate rdfs:label for unit:J-PER-GM-K, and re-serialize.

PR #844, opened 2024-01-05
Add/Fix conversionMultipliers

fkleedorfer:
Checked for errors in conversionMultipliers, or missing ones that can be inferred.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -13727,6 +13727,7 @@ unit:KiloR
   qudt:applicableSystem sou:CGS-EMU ;
   qudt:applicableSystem sou:CGS-GAUSS ;
   qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 0.258 ;
   qudt:hasDimensionVector qkdv:A0E1L0I0M-1H0T1D0 ;
   qudt:hasQuantityKind quantitykind:ElectricChargePerMass ;
   qudt:iec61360Code "0112/2///62720#UAB057" ;
@@ -21479,6 +21480,7 @@ unit:MilliR
   qudt:applicableSystem sou:CGS-EMU ;
   qudt:applicableSystem sou:CGS-GAUSS ;
   qudt:applicableSystem sou:SI ;
+  qudt:conversionMultiplier 2.58e-07 ;
   qudt:hasDimensionVector qkdv:A0E1L0I0M-1H0T1D0 ;
   qudt:hasQuantityKind quantitykind:ElectricChargePerMass ;
   qudt:iec61360Code "0112/2///62720#UAA898" ;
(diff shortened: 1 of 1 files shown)


fkleedorfer:
We can talk about the numeric representation. I can recreate the added triples pretty easily, so if you want a different precision or if you prefer scientific notation, just say so.

(Side glance at #821regarding scientific notation)

fkleedorfer:
Rebased onto latest revision of main

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -31338,3 +31278,1243 @@ THE UCUM TABLE (IN ALL FORMATS), UCUM DEFINITIONS, AND SPECIFICATION ARE PROVIDE
...
+  #       └──┬ unit:J multiplier: 1.0
+  #          ├─── unit:M multiplier: 1.0
+  #          └──┬ unit:N multiplier: 1.0
+  #             ├──┬ unit:KiloGM multiplier: 1.0
+  #             │  └─── unit:GM multiplier: 0.001
+  #             ├─── unit:M multiplier: 1.0
+  #             └─── unit:SEC^-2 multiplier: 1.0 (multiplier^-2: 1.0)
+unit:PER-WB qudt:conversionMultiplier 1.0 .
+
+  # unit:MilliR has conversionMultiplier 2.58E-7, but we've found out that 1 mR = 3.33564095198152049575576714474920E-13 C/kg (relative diff: 1.999994828470326527163271547239499 - big difference)
```
steveraysteveray, line 32338: This one seems wrong to me. Could you confirm?

steveraysteveray: ...and if it is wrong, all the units involving the roentgen need to be adjusted.

steveraysteveray (commented):
Impressive piece of work. I spot-checked some units and it looks good with the exception of the roentgen units. Appreciate you double-checking those - see embedded comment.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
fkleedorfer: I think you are right, good catch!

I've rebased onto main and removed the changes for R, MilliR and KiloR

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -31434,3 +31376,1211 @@ THE UCUM TABLE (IN ALL FORMATS), UCUM DEFINITIONS, AND SPECIFICATION ARE PROVIDE
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "QUDT Unit of Measure Vocabulary Metadata Version 2.1.34" ;
 .
+
+
+
+  # random correction
+unit:MilliR_man
+    qudt:isScalingOf unit:SV ;
+    qudt:conversionMultiplier 0.0001 .
```
steveraysteveray, line 31385: System choked on the trailing .
Should be ;

steveraysteveray (changes requested):
See the one trivial syntax error.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
fkleedorfer: Fixed. And I removed the qudt:isScalingOf triple, as these have been removed from QUDT long ago. Qudtlib still relies on them for building the factor unit hierarchy, so I've got a separate file with these - this triple slipped into the PR by accident.

fkleedorfer:
done

steveraysteveray (approved):

PR #869, opened 2024-03-04
Draft: Add triples for numeric values in xsd:double and xsd:decimal

fkleedorfer:
This allows for using scientific notation, which provides a better developer experience. Equivalence between the values should be checked via SHACL constraints.

* For `qudt:Unit` individuals:
  - qudt:conversionMultplier (xsd:decimal)
  - qudt:conversionMultplierSN (xsd:double)
  - qudt:conversionOffset (xsd:decimal) - if present
  - qudt:conversionOffsetSN (xsd:double) - if present
* For `qudt:ConstantValue` individuals:
  - qudt:value (xsd:decimal)
  - qudt:valueSN (xsd:double)
  - qudt:standardUncertainty (xsd:decimal) - if present
  - qudt:standardUncertaintySN (xsd:double) - if present


fkleedorfer:
So far, the original triples have not be removed, so there may be duplicates.

fkleedorfer:
I just noticed that this is based on the release data, not the latest state of the main branch. Will fix that asap

fkleedorfer:
@steveraysteveray how do you want to remove the original triples? 

a) I can generate a list of triples to be deleted
b) I can write an update query that deletes the triples

Or maybe you have another way using your RDF IDE?

steveraysteveray:
An update query would be great. Thanks for your help.

fkleedorfer:
Here are the queries for deleting the spurious properties:

https://gist.github.com/fkleedorfer/90d8b2b0720f3154bfc367b06e422e65

steveraysteveray (approved):

fkleedorfer:
I'll see if I can adapt the SHACL constraint that checks them for equality from #821

steveraysteveray:
Great! In the end I used variants of the following query to do the cleanup (Note the various commented-out lines). It can also be used with a FILTER != line to test for equality. We found that the SPARQL engine in TopBraid doesn't look any deeper than 16 digits when comparing an xsd:double to an xsd:decimal.
```
#SELECT (COUNT (DISTINCT ?u) AS ?count)
#DELETE {?u qudt:standardUncertainty ?m .}
SELECT *
WHERE {
?u a qudt:ConstantValue .
#?u qudt:standardUncertainty ?m .
#BIND (DATATYPE(?m) AS ?mT) .
#FILTER (?mT = xsd:double) .
?u qudt:standardUncertainty ?n .
BIND (DATATYPE(?n) AS ?nT) .
FILTER (?nT = xsd:decimal) .
?u qudt:standardUncertaintySN ?s .
BIND (DATATYPE(?s) AS ?sT) .
}
```

PR #870, opened 2024-03-08
Add SHACL shapes ensuring correctness of scientific notation values

fkleedorfer:
Constraints added for
- qudt:conversionMultiplier(SN)
- qudt:conversionOffset(SN)
- qudt:value(SN)
- qudt:standardUncertainty(SN)

### DIFF
--- a/collections/COLLECTION_QUDT_QA_TESTS_ALL-v2.1.ttl
+++ b/collections/COLLECTION_QUDT_QA_TESTS_ALL-v2.1.ttl
@@ -413,7 +413,7 @@ qudt:conversionMultiplierSnShape
                             $this
                                 qudt:conversionMultiplier ?valueDecimal ;
                                 qudt:conversionMultiplierSN ?valueSN .
-                            BIND(REPLACE(STR(?valueDecimal), "(\\\\.\\\\d\\\\d*)0+$", "$1") as ?valueDecimalNoTrailingZeros)      # remove trailing zeros from valueDecimal (string)
+                            BIND(REPLACE(STR(?valueDecimal), "(\\\\.\\\\d)(\\\\d*[1-9])?0*$", "$1$2") as ?valueDecimalNoTrailingZeros)      # remove trailing zeros from valueDecimal (string)
                             BIND(REPLACE(STR(?valueSN), "[eE]-?\\\\d+$", "") as ?m)                                           # extract the mantissa (string)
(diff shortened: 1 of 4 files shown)


steveraysteveray (commented):
I added the new properties and property shapes to the OWL and SHACL schema files.

One observation: I tried adding a bunch of zeroes at the end of the decimal value of a ConstantValue, and the validation fails. I have not examined the mechanism of your validation rules - just the behavior. Is there a way to make it pass in that case? I suppose the counter-argument is the implied precision of the trailing zeroes... I'd welcome everybody's thoughts on this.

fkleedorfer:
I actually tried to allow trailing zeroes. There is a case with one trailing zero, which passes, if I am not mistaken... It's one regex that does that. First we have to figure out what behavior we want, then we can change the regex

steveraysteveray:
Agreed. Let's see what others think. I was going to test adding some non-zeroes beyond the resolution of xsd:double to see if it would pass, but didn't get that far. My thought is that a deviation beyond the resolution of xsd:double is fine. We should consider the xsd:decimal as the "master value", and the xsd:double as a human-readable approximation.

ralphtq (commented):
Following discussion with Steve the decimal values should be treated as the definitive values. Should the scientific value be relied on to a specific number of digits, that is, up to 16 digits? The SN value is really intended for human consumption and not for calculation.  Therefore we should not insist on strict matching on trailing zeros. Or any number beyond 16 digits. @fkleedorfer thoughts?

fkleedorfer:
I actually tried to allow trailing zeroes. There is a case with one trailing zero, which passes, if I am not mistaken... It's one regex that does that. First we have to figure out what behavior we want, then we can change the rege

fkleedorfer:
Here is what I think we all want:

1. SN is optional
2. SN is a human-read/writable constraint on the value 
3. Precision is not specified by SN having more or fewer trailing zeroes

Hence, trim trailing zeros behind the decimal point except for the first, in both SN and ctual value before comparing.

steveraysteveray:
I'm a little unclear on your trimming. Would this be how you might trim?

0.00100 trims to 0.001
1.00E-3 trims to 1.0E-3

...and then you compare?

fkleedorfer:
That's what I meant

fkleedorfer:
That's what I meant

fkleedorfer:
Updated the shapes to do exactly what's described above:
1. if there is no SN value: validation passes
2. Any trailing zeroes of the SN value are disregarded in the comparison
3. Any trailing zeroes of the decimal value are disregarded in the comparison

PR #875, opened 2024-03-19
Replace quantitykind:Unknown with relevant quantity kinds for many units

mgberg:
This replaces `quantitykind:Unknown` with relevant quantity kind(s) for ~230 units where possible based on similarity with other units. For example, the quantity kinds specified for `unit:A` were applied to `unit:AttoA`.

Additionally, a few `qudt:exactMatch` statements were added. SHACL validation passed.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -128,7 +128,8 @@ unit:A-HR-PER-DeciM3
   qudt:conversionMultiplier 3600000.0 ;
   qudt:conversionMultiplierSN 3.6E6 ;
   qudt:hasDimensionVector qkdv:A0E1L-3I0M0H0T1D0 ;
-  qudt:hasQuantityKind quantitykind:Unknown ;
+  qudt:hasQuantityKind quantitykind:ElectricChargeDensity ;
+  qudt:hasQuantityKind quantitykind:ElectricChargeVolumeDensity ;
   qudt:iec61360Code "0112/2///62720#UAD883" ;
   qudt:symbol "A·h/dm³" ;
   qudt:ucumCode "A.h.dm-3"^^qudt:UCUMcs ;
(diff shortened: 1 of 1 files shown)


steveraysteveray (approved):
Thanks for running these through the validator!

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -1023,7 +1027,13 @@ unit:AttoA
...
   qudt:conversionMultiplierSN 1.0E-18 ;
   qudt:hasDimensionVector qkdv:A0E1L0I0M0H0T0D0 ;
-  qudt:hasQuantityKind quantitykind:Unknown ;
+  qudt:hasQuantityKind quantitykind:CurrentLinkage ;
+  qudt:hasQuantityKind quantitykind:DisplacementCurrent ;
+  qudt:hasQuantityKind quantitykind:ElectricCurrent ;
+  qudt:hasQuantityKind quantitykind:ElectricCurrentPhasor ;
+  qudt:hasQuantityKind quantitykind:MagneticTension ;
+  qudt:hasQuantityKind quantitykind:MagnetomotiveForce ;
+  qudt:hasQuantityKind quantitykind:TotalCurrent ;
```
dr-shorthair, line 1036: Question: are there any undesirable entailments following from multiple `hasQuantityKind` properties? 
In particular, is it clear that the values are OR, not AND?

steveraysteveray: I don't see any undesirable implications. The triple is basically saying "This unit is a valid unit when quantifying \<quantitykind\>". It is not an OR.

What our "applicableUnit" algorithm does is use all such triples when computing the applicable units for a given QuantityKind, as described [here](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#4-computing-applicable-units-for-a-quantitykind).

Is there something you are worried about?

PR #882, opened 2024-03-25
Added proposal for equivalents

glow-mdsol:
Equivalents are non-SI but commonly used unit in Biological Sciences; this PR adds references for Equivalents and Equivalent Concentrations.


steveraysteveray:
Let me start by thanking you for bringing up what is really a larger issue.

This is a case where I currently disagree with you (and with Wikipedia, it seems) that these are in fact Units. I would argue that they are QuantityKinds, i.e. an identifier for what a measurement is of, that itself is quantified with a Unit. In other words, when you report something in "Equivalents" you still would identify the Unit as moles or whatever is appropriate. So, I would say:
```
quantitykind:EQ skos:broader quantitykind:AmountOfSubstance ;

and

ex:MyMeasurement qudt:hasQuantityKind quantitykind:EQ ;
ex:MyMeasurement qudt:hasUnit unit:MOL ;

and

ex:AnotherMeasurement qudt:hasQuantityKind quantitykind:EQ ;
ex:AnotherMeasurement qudt:hasUnit unit:MilliMOL ;
```
(Similarly for quantitykind:EQPerUnitVolume, which could have skos:broader quantitykind:AmountOfSubstancePerUnitVolume.)

The prefixed **size** of the Unit is already handled in the units vocabulary. All you need to add is the different dimensionality versions of EQ (i.e. EQ, EQPerUnitVolume, etc.)

Finally, to make the applicableUnit algorithm work correctly (you shouldn't explicitly declare the applicable units, as explained [here](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#4-computing-applicable-units-for-a-quantitykind)), you should add the following triples to the Units graph:
```
unit:MOL qudt:hasQuantityKind quantitykind:EQ ;
unit:MilliMOL qudt:hasQuantityKind quantitykind:EQ ;
etc.
```
You can look at the applicable units for [quantitykind:AmountOfSubstance](http://qudt.org/vocab/quantitykind/AmountOfSubstance) and decide which of those units need to also point to quantitykind:EQ.

This way, when the algorithm is run in the next release, quantitykind:EQ will get:
```
quantitykind:EQ qudt:applicableUnit unit:MOL ;
quantitykind:EQ qudt:applicableUnit unit:MilliMOL ;
etc.
```

This approach also works for CO2-Equivalent as a concept, when discussing contributions to climate change, which is reported in Tons, or KiloTons, or whatever is the appropriate unit.

Having said all this, I'm open to discussion. If you do agree, I'd be happy to work with you to get the relationships right.

glow-mdsol:
Sure, all fair points.  

I hadn't thought about these as discrete quantity kinds themselves.   We see this primarily for Blood Chemistry results, such as CO2 concentration, Bicarbonate Concentration, Chloride Concentration, Magnesium Concentration. 
Our consumers are receiving lab results using mEq/L (as in this example) and need convert to mol/L (the SI unit in this case).  I'll make some updates and push tomorrow

steveraysteveray:
The joy and the agony of using an ontology is that it forces people to be precise in their use of units. When your consumers are receiving lab results that say "Chloride concentration is 34.3 mEq/L", what is really meant is the number is a measure of the Equivalent for Chloride and the **unit of measure** is milli-mole per litre of the reaction unit (presumably H+). The magnitude, and thus the conversion to mol/L of Chloride, is not a fixed number because it depends on the valence of Chloride in the context of the (implicit) chemical reaction being talked about. Is that correct?

steveraysteveray:
Great. Let me know if you need help on your PR.

steveraysteveray (changes requested):
A comment, some needed changes, plus a request for how you submit PRs in the future.

**Comment**: I had suggested that you don't really need new units, just the new quantity kinds. While that is strictly speaking correct, I can understand that people might be confused if they see a measured unit in moles, and don't notice that the measurement has a quantity kind of Equivalent. So in the spirit of clarity, I can go along with your two new units.

**Needed changes**: Various triples are missing, shown in this screen shot.
image

**A request**: You followed the [submission guideline](https://github.com/qudt/qudt-public-repo/wiki/Unit-Vocabulary-Submission-Guidelines#adding-a-unit-to-an-existing-vocabulary-file-recommended-method) by creating the new file in the submissions folder, but it is actually much easier for our review if you just edit the vocabulary files directly in your branch. (I kind of regret offering the other option!). So for this PR, I'm happy to copy the new definitions over to their respective graphs, but could you just edit the unit etc. files directly in the future? It would be most appreciated.

steveraysteveray:
@glow-mdsol, after some research and discussion with colleagues, I'd like to step back to my earlier [comment on March 26](https://github.com/qudt/qudt-public-repo/pull/882#issuecomment-2020824019). Namely, creating units for reporting measurements of Equivalents is a deep hole to descend into because one would need to create units of many sizes (milli, micro, nano, etc) for each of the quantity kind versions of quantitykind:Equivalent (mass, molar, concentration, density), resulting in an unwieldy number of units to maintain.

As I stated in my earlier comment, there really is no need for special units - the normal set of units such as MilliMole (unit:MilliMOL), MilliMole per Litre (unit:MilliMOL-PER-L), etc. are precise and correct. 

What we will do to meet your needs is to define the following:

A quantity kind for each dimensionality:
1. Molar amount (**quantitykind:Equivalent_Molar**), with dimension vector qkdv:A1E0L0I0M0H0T0D0
2. Mass amount (**quantitykind:Equivalent_Mass**), with dimension vector qkdv:A0E0L0I0M1H0T0D0
		Note that the currently defined quantitykind:CO2Equivalent will have a skos:broader relationship to quantitykind:Equivalent_Mass
3. Molar-based **quantitykind:EquivalentConcentration**, with dimension vector qkdv:A1E0L-3I0M0H0T0D0
4. Mass-based **quantitykind:EquivalentDensity**, with dimension vector qkdv:A0E0L-3I0M1H0T0D0

Each of these four quantity kinds will have a skos:broader relationship with their "non-equivalent" counterparts, namely
1. quantitykind:AmountOfSubstance
2. quantitykind:Mass
3. quantitykind:Concentration
4. quantitykind:Density

And as stated in the earlier comment, your own defined measurement (aka instance of qudt:Quantity) would point to the appropriate quantity kind (such as quantitykind:EquivalentConcentration), and point to the relevant unit, (such as unit:MilliMOL-PER-L). Your measurement would need to identify what chemical species the measurement is for, which is outside the scope of QUDT. I can imagine something like the following:

```
ex:MyMeasurement a qudt:Quantity .
ex:MyMeasurement qudt:hasQuantityKind quantitykind:EquivalentConcentration .
ex:MyMeasurement qudt:hasUnit unit:MilliMOL-PER-L .
ex:MyMeasurement myChemistryOntology:ofChemicalSpecies myChemistryOntology:CalciumIon .
```

We believe the above approach will meet your needs. Please stay tuned for PR #931 that will establish the above definitions.

glow-mdsol:
This is a great solution.   Thanks for your patience with this,  we are definitely going to need to have some way to incorporate the valence on our side for unit conversions.

steveraysteveray:
Glad that you feel this works for you. I'll go ahead and close this PR.

PR #887, opened 2024-03-27
Removed "kg" as udunit for PINT_US-PER-DAY

glow-mdsol:
- I think this is a copy-paste error - I can't find this unit anywhere in the upstream and kg only matches KiloGM


steveraysteveray (approved):
Thanks for catching this.

PR #893, opened 2024-04-09
Added units

gezever:
Added units to publish the flow of pollution caused by factories on a yearly base, as required by EU-laws.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -13244,7 +13244,9 @@ unit:HR-PER-YR
     qudt:applicableSystem sou:USCS ;
     qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
     qudt:hasQuantityKind quantitykind:TimeRatio;
-    qudt:plainTextDescription "hours (of activity) per year  " ;
+    qudt:qkdvDenominator qkdv:A0E0L0I0M0H0T1D0 ;
+    qudt:qkdvNumerator qkdv:A0E0L0I0M0H0T1D0 ;
+    qudt:plainTextDescription "hours (of activity) per year" ;
     qudt:symbol "hr/yr" ;
     qudt:symbol "hr/jaar" ;
     qudt:ucumCode "hr/a"^^qudt:UCUMcs ;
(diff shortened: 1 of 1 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -13234,6 +13234,25 @@ unit:HR-PER-FT2
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "hour per square foot" ;
 .
+unit:HR-PER-YR
```
steveraysteveray, line 13237: To be even more sophisticated, consider adding a numerator and denominator dimension vector (each with dimension of time) as described [here](https://github.com/qudt/qudt-public-repo/wiki/User-Guide-for-QUDT#6-using-quantity-kinds-with-dimensionless-quantities).

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -23537,6 +23557,25 @@ unit:M3-PER-SEC2
...
   rdfs:label "Cubic Metre per Square Second"@en ;
 .
+unit:M3-PER-YR
+    a qudt:Unit ;
+    dcterms:description "power of the SI base unit metre with the exponent 3 divided by the unit year"^^rdf:HTML ;
+    qudt:applicableSystem sou:CGS ;
+    qudt:applicableSystem sou:CGS-EMU ;
+    qudt:applicableSystem sou:CGS-GAUSS ;
+    qudt:applicableSystem sou:SI ;
+    qudt:conversionMultiplier 3.168808781e-8 ;
```
steveraysteveray, line 23567: Should be 
qudt:conversionMultiplier 0.00000003168808781 ;
qudt:conversionMultiplierSN 3.168808781E-8 ;

steveraysteveray (changes requested):

gezever:
@steveraysteveray 
I committed the suggested corrections.

steveraysteveray (approved):

gezever:
@steveraysteveray 
Thank You

steveraysteveray:
@fkleedorfer, you are correct that two of the three units use "year" rather than "yr", which should be corrected, although we do not have a validation rule or algorithm to ensure consistency.

More broadly though, the multiple symbols raise some options worth discussing:

Should we:
1. Disallow multiple symbols
2. Allow multiple symbols, each with a different language tag (like we do with rdfs:label)
3. Allow multiple symbols, possibly with the same language tag

These three new units are the only ones with more than one symbol at present. However, it's possible there there exist some units with alternate symbols, yes?
Currently we do not validate against any of the three options above. Should we?

We do want symbols to be represented using Unicode.

fkleedorfer:
I'm on team 1. we thus avoid the problem of having to choose one, and we have a clear path toward automatic symbol generation/verification

fkleedorfer:
I think there is a lot more to say about this topic, so I started a [discussion about it](https://github.com/qudt/qudt-public-repo/discussions/895)

Issue #929, opened 2024-05-07
SpecificEnthalpy missing as kind in related SpecificEnergies


ohshazbot:
The current corpus of [SpecificEnthalpy](https://qudt.org/vocab/quantitykind/SpecificEnthalpy) is a bit lacking, as only 2 units currently map even though there are several matching units in the very similar kind SpecificEnergy. I could be missing some details about how qudt is approaching enthalpy, but my git searches couldn't find anything. So I think it's just a matter of adding the missed kind labels, but this could be symptom of a missed abstraction?

If it's the former at least, I'll throw up a PR

steveraysteveray:
Great observation. Actually, your solution in #930 is indeed a reasonable stopgap solution, but it would be better to go ahead and solve the more general problem. The way we generate applicableUnits is described [here](https://github.com/qudt/qudt-public-repo/wiki/Advanced-User-Guide#4-computing-applicable-units-for-a-quantitykind). So the best way to get all the applicableUnits listed for SpecificEnthalpy would be to fix our skos:broader relationships. We could then have the various units point to quantitykind:SpecificEnergy via the qudt:hasQuantityKind relation, and the narrower quantity kinds would "inherit" the applicableUnits as per the algorithm described.

It's not a trivial exercise, because some quantity kinds with the same dimension vector are best used with different, non-energy related units. I'm thinking specifically about quantity kinds like quantitykind:MassicTorque and quantitykind:SpecificModulus. Here's the listing of all the quantity kinds with the same dimension vector:
![image](https://github.com/qudt/qudt-public-repo/assets/1130189/f24c68d7-d9a3-478e-ad67-ebe06c6a478b)

So, the solution involves constructing the correct skos:broader hierarchy involving some (but not all) of these quantity kinds - that is, all the energy-related ones under SpecificEnergy, but not the force or dose related ones.

 Does this make sense? Would you be interested in taking a crack at this? If not, let me know and I'll put it on my list!

ohshazbot:
Whelp, you got to it before I could (probably wouldn't have been for a few weeks for me).

For my own clarification though, I'm not fully understanding all of the ```qudt:applicableUnit``` changes in #932. I'm guessing that's mostly cleanup to align things, but does that mean qudt is doing double bookkeeping for tracking units<->kinds relationships?

steveraysteveray:
[...]

PR #930, opened 2024-05-07
#929 Adding SpecificEnthalpy to BTU/lb, CAL/G, MJ/kG, etc

ohshazbot:
As mentioned in https://github.com/qudt/qudt-public-repo/issues/929, this may not be the right bigger picture fix, but does address some misses for the timebeing


steveraysteveray:
Resolved by PR #932

Issue #933, opened 2024-05-23
Rate of Change


gwemon:
Hi, I noticed while checking for quantity kinds in QUDT, that there seems to be a duplicate entry for "Rate of Change of Temperature" (http://qudt.org/vocab/quantitykind/RateOfChangeOfTemperature) and "Temperature Rate of Change" (http://qudt.org/vocab/quantitykind/TemperatureRateOfChange). Many thanks, Gwen.

steveraysteveray:
Thanks for spotting this. http://qudt.org/vocab/quantitykind/RateOfChangeOfTemperature was a recent addition, which will be merged with http://qudt.org/vocab/quantitykind/TemperatureRateOfChange.

PR #934, opened 2024-05-23
 Merged rate of change of temperature entries

steveraysteveray:
Addresses #933


vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -20711,15 +20711,6 @@ quantitykind:RankineTemperature
   rdfs:label "Rankine temperature" ;
   rdfs:label "Rankine temperature"@en-US ;
 .
-quantitykind:RateOfChangeOfTemperature
```
fkleedorfer, line 20714: I am personally ok with just removing this, but don't we have the policy to first deprecate, later remove?

steveraysteveray:
Yes, we do. However, the redundant quantity kind just appeared a couple of months ago as part of the IEC harmonization, so (in my judgement) it hasn't been around long enough to be used by anybody. But in general, you are correct.

PR #939, opened 2024-06-06
 Added Standard Cubic Foot and similar units

### LINKED ISSUE #936: Missing unit: SM3-PER-HOUR
We are missing the unit SM3-PER-HOUR (Standard m3/h), like the unit defined here: https://en.wikipedia.org/wiki/Standard_cubic_foot.
Could this be added?

steveraysteveray:
Resolves #936.

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -40833,6 +40833,107 @@ unit:SAMPLE-PER-SEC
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Sample per second"@en ;
 .
+unit:SCF
+  a qudt:Unit ;
+  dcterms:description """
+The $\\textit{standard cubic foot}$ (scf) is a unit representing the amount of gas (such as natural gas) 
+contained in a volume of one cubic foot at reference temperature and pressure conditions. As such, it is a measure of the actual amount of gas,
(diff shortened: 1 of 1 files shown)


steveraysteveray:
Note that I did not use a unit URI like unit:FT3 for Standard Cubic Foot as that would mislead users (and unit name parsers) into thinking it is a unit of volume.

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -40833,6 +40833,107 @@ unit:SAMPLE-PER-SEC
...
+contained in a volume of one cubic foot at reference temperature and pressure conditions. As such, it is a measure of the actual amount of gas,
+not the volume of the gas. The reference conditions for standard cubic foot per minute are 60 degrees Fahrenheit and 14.7 pounds per 
+square inch (psi) of pressure.
+  """^^qudt:LatexString ;
+  qudt:conversionMultiplier 0.0003328 ;
+  qudt:conversionMultiplierSN 3.328E-4 ;
+  qudt:hasDimensionVector qkdv:A1E0L0I0M0H0T-1D0 ;
+  qudt:hasQuantityKind quantitykind:MolarFlowRate ;
+  qudt:informativeReference "https://en.wikipedia.org/wiki/Standard_cubic_feet_per_minute"^^xsd:anyURI ;
+  qudt:symbol "scfh" ;
```
fkleedorfer, line 40865: I would have generated `qudt:symbol "scf/h" ;

fkleedorfer: I should have checked before commenting. It's the correct symbol, of course.

fkleedorfer (approved):

PR #945, opened 2024-07-08
New unit: Swiss Francs per kilogram

tpluscode:
I would like to propose a new unit which will be used at [Lindas Linked Data Services](https://lindas.admin.ch/) to publish Swiss open data

### DIFF
--- a/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
+++ b/vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
@@ -8368,6 +8368,7 @@ unit:CHF-PER-KiloGM
   qudt:applicableSystem sou:SI ;
   qudt:conversionMultiplier 0.001 ;
   qudt:conversionMultiplierSN 1.0 ;
+  qudt:hasQuantityKind quantitykind:CostPerMass ;
   qudt:symbol "CHF/kg" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Swiss Francs per kilogram"@en ;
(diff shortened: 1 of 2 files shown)


vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -8370,7 +8370,7 @@ unit:CHF-PER-KiloGM
   qudt:conversionMultiplierSN 1.0E-3 ;
```
steveraysteveray, line 8370: Since the kilogram is the base SI unit, the conversion multipliers should be 1.0

steveraysteveray (changes requested):
One item that is still missing is the quantity kind CostPerMass that your unit can then point to.

tpluscode:
> One item that is still missing is the quantity kind CostPerMass that your unit can then point to.

Shall I add it in the PR?

vocab/unit/VOCAB_QUDT-UNITS-ALL-v2.1.ttl
```diff
@@ -8367,7 +8367,8 @@ unit:CHF-PER-KiloGM
   a qudt:Unit ;
   qudt:applicableSystem sou:SI ;
   qudt:conversionMultiplier 0.001 ;
```
steveraysteveray, line 8369: This one needs to be 1.0
The SN one needs to be 1.0E0

steveraysteveray (changes requested):
The quantity kind and unit need triples for qudt:hasDimensionVector
Namely:
  qudt:hasDimensionVector qkdv:A0E0L0I0M-1H0T0D0 ;

Also, a brief triple for qudt:plainTextDescription.

See https://github.com/qudt/qudt-public-repo/wiki/QuantityKind-Vocabulary-Submission-Guidelines

tpluscode:
Thank you for your pointers @steveraysteveray

steveraysteveray (approved):

PR #947, opened 2024-07-19
Change a unit from having quantity kind Surface Tension to a broader quantity kind

mgberg:
`unit:LB_F-PER-IN` had quantity kind `quantitykind:SurfaceTension`, which causes a problem because it was the only unit that had that quantity kind. `quantitykind:SurfaceTension` was inheriting all its applicable units from the broader quantity kind `quantitykind:EnergyPerArea` and now it only has that one as applicable. This changes `unit:LB_F-PER-IN` to have quantity kind `quantitykind:EnergyPerArea` to fix the issue.

Whether or not it is the most appropriate for `unit:LB_F-PER-IN` to have quantity kind `quantitykind:EnergyPerArea` is another question, but I didn't want to remove it now if someone else was using that way.


steveraysteveray (approved):
I agree - this change expands the number of applicable units beyond just the Imperial one for SurfaceTension

PR #963, opened 2024-08-29
Rh latex fixes cusp4

ralphtq:
Fixes to LaTeX errors on Units, Quantity Kinds and Constants


steveraysteveray (approved):
I see that you changed the datatype from rdf:HTML to qudt:LatexString for many dcterms:descriptions of quantity kinds, even though they contained no Latex. However, it doesn't do any damage, so I suppose that's fine.

ralphtq:
I made the qudt:LatexString changes in anticipation of more content later. Also rdf:HTML was inappropriate because there was no HTML markup in these strings.

PR #968, opened 2024-09-11
Adding azimuth and tilt quantitykinds

lazlop:
First PR! 
Please let me know if I made any mistakes @steveraysteveray


steveraysteveray (approved):
Looks pretty good! I think I mentioned that the applicableUnit triples will be regenerated with each Release, so you could have skipped them for now, but it's fine the way it is.

PR #971, opened 2024-09-13
 Fixed quoted URIs as replacements

steveraysteveray:


steveraysteveray:
Also fixed a problem related to referencing deprecated units.

jhodgesatmb (approved):
Those quoted URIs look like what happened when I added them manually. I recall mentioning to SR that adding the units TBCME said they were strings even though they were valid units. Some of them. It was a mystery then and still but I didn't think that TBCME would add quotes so I didn't look.

PR #977, opened 2024-10-02
Fix datatype of qudt:currencyNumber

### LINKED ISSUE #973: Currency numbers should be strings, not integer literals
in vocab/currency/VOCAB_QUDT-UNITS-CURRENCY-v2.1.ttl.  the `qudt:currencyNumber` is an integer literal, which is not a good way of representing it. The currency number is a 3-digit code, which is left-filled with zeros if necessary. 

e.g
```
cur:AMD
  a qudt:CurrencyUnit ;
  qudt:currencyCode "AMD" ;
  qudt:currencyExponent 0 ;
  qudt:currencyNumber 051 ;
```

It might (will) happen that this value, `051` in this case,  gets picked up as `51` by a client, which will lead to a bug.

I think we should change it to a string and check it is all digits using SHACL.

- [ ] update data
- [ ] write SHACL
- [ ] update schema
- [ ] enforce using OWL

fkleedorfer:
Change the type from xsd:integer to xsd:string and enforce the pattern "^\\d{3}$" (three digits).

Fixes #973

### DIFF
--- a/collections/COLLECTION_QUDT_QA_TESTS_ALL-v2.1.ttl
+++ b/collections/COLLECTION_QUDT_QA_TESTS_ALL-v2.1.ttl
@@ -19,6 +19,8 @@
 @prefix voag: <http://voag.linkedmodel.org/schema/voag#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
+@prefix cur: <http://qudt.org/vocab/currency/> .
+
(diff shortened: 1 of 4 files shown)


fkleedorfer:
If someone could review the changes to the OWL schema, that would be helpful.

jhodgesatmb (approved):

PR #980, opened 2024-10-08
 Moved unused files to private repo

steveraysteveray:


vocab/scales/VOCAB_QUDT-SCALES-v2.0.ttl
ralphtq, line 1: Should we keep scales?

vocab/quantitykinds/to-be-deprecated-quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ISQ-CHARACTERISTIC-NUMBERS-v2.0.ttl
ralphtq, line 1: Should we keep this?

ralphtq (commented):
Maybe keep characteristic numbers and scales

vocab/scales/VOCAB_QUDT-SCALES-v2.0.ttl
steveraysteveray: No other file in the public repo refers to any of the scale types in this file, so I think it's safe to move it to the private repo, where it can be brought back whenever we want.

vocab/quantitykinds/to-be-deprecated-quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ISQ-CHARACTERISTIC-NUMBERS-v2.0.ttl
steveraysteveray: The characteristic numbers defined in this graph would need quite a bit of work to bring them up to satisfying our validation constraints - they lack dimension vectors, they have skos:exactMatch relations to web page addresses, and much more. At the same time, many of the entries already exist (with all the necessary properties) in the quantity kind graph (e.g. ReynoldsNumber). For that reason, I suggest we move this one to the private repo.

fkleedorfer (approved):

PR #990, opened 2024-10-29
 Fixed most qks lacking applicableUnits

steveraysteveray:
Got to G so far...

### DIFF
--- a/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
+++ b/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
@@ -132,6 +132,7 @@ quantitykind:AbsoluteTypographicMeasurement
   qudt:iec61360Code "0112/2///62720#UAD372" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "absolute typographic measurement" ;
+  skos:broader quantitykind:Distance ;
 .
 quantitykind:AbsorbedDose
   a qudt:QuantityKind ;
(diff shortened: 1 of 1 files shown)


vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
```diff
@@ -132,6 +132,7 @@ quantitykind:AbsoluteTypographicMeasurement
   qudt:iec61360Code "0112/2///62720#UAD372" ;
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "absolute typographic measurement" ;
+  skos:broader quantitykind:Distance ;
```
fkleedorfer, line 135: This is the only occurrence of AbsoluteTypographicMeasurement in the repo, so it is not easy to decide by example, but I believe the correct broader qk would be qk:Length

fkleedorfer (commented):
If the point of this PR was to add a skos:broader relation to all qks that did not have one: there are a few that still need to be done. Apart from the comment, the added triples seem fine to me

vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl
steveraysteveray: Agreed! For some reason I had read "Topographic" and concluded there must be a datum involved, as with Distance.
Now fixed.

steveraysteveray:
@fkleedorfer, there are indeed 76 remaining QuantityKinds. I left them off this PR because some are quite complex and need some research. Others seem to me to have erroneous dimension vectors (odd things like GasLeakRate having a dimension consistent with Power; or SpecificThrust that has a description implying specific impulse, which is not the same as thrust (i.e. force). Each of these will take some time, so I thought I would stop with the "easy" ones for now.

steveraysteveray:
To your original question, there is of course nothing wrong with a quantity kind not having a skos:broader. I was chasing cases lacking applicableUnits, and using skos:broader is an easy way to produce applicableUnits. It's a bit crude of a fix, I admit, but better than having none at all.

fkleedorfer (approved):

PR #993, opened 2024-11-07
Replace qudt:hasDimensionVector with qudt:dimensions in datatypes file

### LINKED ISSUE #992: Use of qudt:hasDimensionVector in datatypes file clashes with other uses
The `qudt:hasDimensionVector` here is wrong, I believe:
```
qudt:MATRIX-4X4-FLOATDP
  a qudt:Matrix ;
  qudt:code 320 ;
  qudt:dimensionality 2 ;
  qudt:elementType qudt:FLOAT-DP ;
  qudt:hasDimensionVector [
      a qudt:DimensionVector ;
      rdf:first "4"^^xsd:int ;
      rdf:rest (
          "4"^^xsd:int
        ) ;
      rdfs:label "4 by 4 array" ;
    ] ;
  qudt:typeMatrix <http://qudt.org/schema/type#TYPE-MATRIX-4x4-FLOATDP> ;
  rdfs:isDefinedBy <http://qudt.org/2.1/vocab/datatype> ;
 [...]
```
I don't really know what this definition is used for (if anything), but it trips up the turtle-formatter. 
It seems that it is an accident to call it 'dimensionVector' - which, in its widely used meaning denotes the SI dimension vector. Here, it seems to denote the dimensions of a matrix

How about we change it to

```
qudt:dimensions (4 4);
```

fkleedorfer:
whila also simplifying the dimension lists where appropriate.

Fixes #992

### DIFF
--- a/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
+++ b/schema/SCHEMA_QUDT-DATATYPE-v2.1.ttl
@@ -159,11 +159,6 @@ qudt:Array
     ] ;
   rdfs:subClassOf [
       a owl:Restriction ;
-      owl:allValuesFrom qudt:DimensionVector ;
-      owl:onProperty qudt:dimensionVector ;
-    ] ;
-  rdfs:subClassOf [
-      a owl:Restriction ;
       owl:allValuesFrom <http://qudt.org/schema/type#TypeMatrix> ;
       owl:onProperty qudt:typeMatrix ;
     ] ;
@@ -180,7 +175,7 @@ qudt:Array
   rdfs:subClassOf [
       a owl:Restriction ;
       owl:maxCardinality 1 ;
-      owl:onProperty qudt:dimensionVector ;
+      owl:onProperty qudt:dimensions ;
     ] ;
   rdfs:subClassOf [
       a owl:Restriction ;
@@ -243,7 +238,7 @@ qudt:ArrayType
   rdfs:subClassOf qudt:CompositeDatatype ;
   sh:property qudt:ArrayType-byRow ;
   sh:property qudt:ArrayType-dataOrder ;
-  sh:property qudt:ArrayType-dimensionVector ;
+  sh:property qudt:ArrayType-dimensions ;
   sh:property qudt:ArrayType-dimensionality ;
   sh:property qudt:ArrayType-typeMatrix ;
 .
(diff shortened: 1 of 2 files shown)


ralphtq (commented):
The use of qudt:dimesions (n1 n2) for 2 dimensional matrices works, qudt:dimenssion (n1 n2 n3) for 3 dimensional matrices will also work, etc.

We need support in the schema SCHEMA_QUDT-DATATYPE-v2.1.ttl for qudt:dimensions.  I will do this  as an update to this branch.

ralphtq:
I've reviewed https://github.com/qudt/qudt-public-repo/pull/993 and updated the Datatypes SCHEMA

fkleedorfer:
Looks like we are good to go, but we need @steveraysteveray or @jhodgesatmb to approve this PR now.

steveraysteveray (approved):
I will point out that this file is a mix of SHACL and pure OWL classes. It is not part of our imports closure graph, so it does not get used when following imports. It is validated by @fkleedorfer's code because it is in the schema folder. 

Bottom line, I'm fine with the fix, but eventually we should decide to either make distinct SHACL and OWL versions of this file, or not include it in the repo.

steveraysteveray:
I decided to run the validator on the SCHEMA_QUDT-DATATYPE-v2.1.ttl file, with the following results. Again, this file is not normally validated because it is not imported, but it IS included in our Releases. 
Screenshot 2024-11-08 at 9 29 43 AM

We can keep it in the Release if you want, or we can move it aside until we fix the various issues.

ralphtq:
I intend to fix the issues. Some refactoring into the QUDT schema may be warranted.Ralph HodgsonOn Nov 8, 2024, at 10:03 AM, steveraysteveray ***@***.***> wrote:﻿
I decided to run the validator on the SCHEMA_QUDT-DATATYPE-v2.1.ttl file, with the following results. Again, this file is not normally validated because it is not imported, but it IS included in our Releases.
Screenshot.2024-11-08.at.9.29.43.AM.png (view on web)
We can keep it in the Release if you want, or we can move it aside until we fix the various issues.

steveraysteveray:
@ralphtq, will you be creating SHACL and OWL versions? Again, we aren't referencing anything in this graph from our current vocabularies, so it is not urgent that we fix this if we don't include it in the Release.

PR #1004, opened 2024-11-14
 Updated documentation for the Tesla

steveraysteveray:


jhodgesatmb (approved):

fkleedorfer (approved):

fkleedorfer:
We should have updated the README.md file there 

As we can see in other PRs these days, not every PR warrants it, but maybe we should add a check that displays a warning in the PR dialogue if that is possible

PR #1012, opened 2024-11-14
Change SHACL shapes to enforce only one symbol per entity

### LINKED ISSUE #1011: SHACL shapes do not enforce one qudt:symbol per entity

fkleedorfer:
This PR is not ready to merge.

It reveals that several quantitykinds have 2 symbols, we need to fix this before merging

Fixes #1011


fkleedorfer:
have a look at the failing checks to see which quantitykinds need to be fixed

fkleedorfer (commented):
An entry in the changelog would be good

steveraysteveray:
OK, I made my best guesses on what should be the qudt:symbol value, and what should be the qudt:altSymbol value.
This is now ready for review and merge.

pom.xml
```diff
@@ -135,11 +144,9 @@
                                     <!--skip>true</skip-->
                                     <shapes>
                                         <include>target/dist/validation/COLLECTION_QUDT_QA_TESTS_ALL.ttl</include>
-                                        <include>target/dist/schema/shacl/SCHEMA_QUDT_NoOWL.ttl</include>
```
fkleedorfer: We lost two lines in pom.xml by copying the pom.xml from the main branch - the lines that actually cause the cardinalities to be checked. 

Rebase would have been the way to go

fkleedorfer:
Good to go now!

steveraysteveray (approved):

PR #1034, opened 2024-11-22
Correct formatting

fkleedorfer:
The last merge into main included a file that was not formatted properly. 

Seems like the build does not get triggered by a rebase and force-push that happens when you click 'Update Project' in the PR GUI... Possibly that's a different event we need to include in the triggers for the maven CI action


steveraysteveray (approved):

fkleedorfer:
The snapshot release was not updated because the build had failed because of formatting errors

PR #1038, opened 2024-11-26
Fix broken changelog body generation

fkleedorfer:
- the footer was overwriting anything that was added before
- also, the release footer does not fit so well for the snapshot release, so the content was removed.


fkleedorfer:
Note: I just deleted the snapshot release manually so the PR merge will create a new one, which will be the situation of the first PR after a release (because we now delete the snapshot release while releasing).

steveraysteveray (approved):

Issue #1043, opened 2024-11-26
QK for IfcCompoundPlaneAngleMeasure - plane angle measured in degrees, minutes and seconds?


fkleedorfer:
I am considering mappin a [IfcCompoundPlaneAngleMeasure](https://standards.buildingsmart.org/IFC/DEV/IFC4_2/FINAL/HTML/schema/ifcmeasureresource/lexical/ifccompoundplaneanglemeasure.htm) to QUDT. 

So far, mapping QUDT quantityKinds to IFC measures has worked pretty well, but in this case, the IFC measure is about the representation of the measurement rather than about how it is obtained (IFC has a IfcPlaneAngleMeasure). 

The IFC specification is a 4-tuple of integer numbers (degrees, minutes, seconds, millionths of second). 

I would like to know how one would represent such a value in QUDT (lesser priority though), and whether it would be permissible to add this quantitykind (`qk:CompoundPlaneAngleMeasure`)- if not for any other use than making a 1:1 mapping between QUDT and IFC a bit more feasible. (higher priority)

BTW, I have no ideas yet on how to integrate this with the conversionFactor/offset system. Interested in opinions as well.

jhodgesatmb:
Measurements, sensors, and actuators are best handled by SOSA. In our EPIC project we build bridges between QUDT and SOSA and SOSA to IFC. I may be mistaken as I have constructed many bridge ontologies. I’ll look them up and send a diagram.The long and short of it is that if an ontology is purely about units and quantities we have suggested a mapping/harmonization, if it is a taxonomy we have suggested a cross reference, and if it is a domain we have suggested a bridge. Perhaps it is just my recommendation but I am a believer that ontologies should stand on their own and be integrated pairwise with other ontologies to increase and simplify interoperability.Jack Hodges, Ph.D.Arbor StudiosOn Nov 26, 2024, at 6:24 AM, Florian Kleedorfer ***@***.***> wrote:﻿
I am considering mapping a IfcCompoundPlaneAngleMeasure to QUDT.
So far, mapping QUDT quantityKinds to IFC measures has worked pretty well, but in this case, the IFC measure is about the representation of the measurement rather than about how it is obtained (IFC has a IfcPlaneAngleMeasure).
The IFC specification is a 4-tuple of integer numbers (degrees, minutes, seconds, millionths of second).
[...]

PR #1048, opened 2024-12-02
Contribute Quantitykinds for frequency and angle, fix unit localnames

fkleedorfer:
### Adds 
- `qk:CompoundPlaneAngle` (#1043)
- `qk:CountRate` - needed to map [IfcIntegerCountRateMeasure](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2/HTML/link/ifcintegercountratemeasure.htm)
- `qk:RotationalFrequency` - needed to map [IfcRotationalFrequencyMeasure](https://standards.buildingsmart.org/IFC/RELEASE/IFC4/ADD2/HTML/link/ifcrotationalfrequencymeasure.htm)

### Fixes
- Renamed incorrectly named Unit `unit:CI-PER-KiloGM` to `unit:Ci-PER-KiloGM`
- Renamed incorrectly named Unit `unit:Flight` to `unit:FLIGHT` (determined wrong because there are derived units using FLIGHT not Flight)


src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -5857,6 +5859,18 @@ unit:Ci
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/unit> ;
   rdfs:label "Curie"@en .
 
+unit:Ci-PER-KiloGM
```
steveraysteveray, line 5862: Our naming convention for units is all uppercase, except for prefixes, so it should be CI

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
```diff
@@ -11685,6 +11707,18 @@ quantitykind:RotaryShock
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "rotary shock"@en-US .
 
+quantitykind:RotationalFrequency
```
steveraysteveray, line 11710: How is this different from AngularVelocity that already exists?

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
```diff
@@ -2276,6 +2286,18 @@ quantitykind:Count
   rdfs:label "Count"@en ;
   skos:broader quantitykind:Dimensionless .
 
+quantitykind:CountRate
```
steveraysteveray, line 2289: We already have the skos:broader hierarchy with InverseTime at the top, and which includes Incidence. Are you thinking we need this as well?

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
fkleedorfer: ah ok

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
fkleedorfer: I don't know where that was coming from. I had already identified the qk I need as  `qudt:AngularFrequency`. Never mind. Will remove from PR.

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
fkleedorfer: I do. Incidence is apparently from epidemiology, this is more general. Could it be a broader kind of incidence?

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
fkleedorfer: No, actually, It seems we need RotationalFrequency: https://en.wikipedia.org/wiki/Rotational_frequency - this description explains the distiction between it and AngularVelocity/AngularFrequency and maps nicely to http://standards.buildingsmart.org/IFC/DEV/IFC4_2/FINAL/HTML/link/ifcrotationalfrequencymeasure.htm

fkleedorfer: ... and I would also like to suggest new units for RPM and cycles per second (that's the one they use in IFC)

fkleedorfer: AngularFrequency and AngularVelocity confuse me. 

The suggested distinction of RotationalFrequency and RotationalVelocity is quite subtle, but probably ok.

my suggestion about RPM was an error. I ended up finding REV-PER-MIN, and that one is fine.

fkleedorfer: Looking at https://en.wikipedia.org/wiki/Rotational_frequency it seems that Hz, cycles per time and revolutions per time would all be units of rotational frequecy. I don't know what we would need rotational velocity for or what units it should be associated with.

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
```diff
@@ -6158,11 +6180,11 @@ quantitykind:IonConcentration
   a qudt:QuantityKind ;
   dcterms:description "\"Ion Concentration\" is the number of ions per unit volume. Also known as ion density."^^qudt:LatexString ;
   qudt:exactMatch quantitykind:IonDensity ;
-  qudt:hasDimensionVector qkdv:A0E0L-3I0M0H0T0D0 ;
+  qudt:hasDimensionVector qkdv:A1E0L-3I0M0H0T0D0 ;
```
steveraysteveray, line 6183: This change opens up a larger discussion: Is a numeric count dimensionally the same as "amount of material" measured in moles? 

Argument in favor: Yes, you can just choose which unit to use, be it a mole or a number
Argument against: No, a number or count is a dimensionless concept, not the SI concept of "amount of material" that is dimensionally encoded as A1 in our dimension vector.

In the past I have considered number as a dimensionless concept.

Thoughts?

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
steveraysteveray: OK, I propose we make RotationalFrequency and RotationalVelocity qudt:exactMatch for now. We could refine this once we fully support vector datatypes. The applicable units would include Hertz, cycles per time, and revolutions per time.

Also, that we make AngularFrequency and AngularVelocity qudt:exactMatch.  The applicable units would include angular units per time (e.g. radians per second), but not Hertz.

Finally, I don't really see the Rotational and Angular quantity kinds with either family clearly skos:broader of the other, so I propose we simply cross reference them with rdfs:seeAlso.

fkleedorfer: i'll do exactly that.

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
fkleedorfer: Looking through the Web, I think the term 'ion concentration' is not sufficiently defined. It refers to the amount of a substance in a solvent (in chemistry), and might either be SI Amount (mol) or a relative count (e.g. ppm, which is mg/L), and there is Molality (did not delve into that) and Normality (neither). 

Then there is ion density (exact match in qudt), which does not seem to be an exact match at all: it is either number density (#/m3), mass density(kg/m3) or charge density(coulomb/m3)

So for our system, in which a quantitykind is associated with one dimension vector, we need quite a bunch of them to cover ion concentration and ion density.

I touched it because IFC thinks IonConcentration is g/L, which until now I thought was wrong, but now (thanks to your watchful eye) I believe it is actually correct, and I think I need to add at least that special kind.

If we wanted to be more correct than we are currently, we may consider
 
IonAmountConcentration
IonCountConcentration
IonMassConcentration (that's the one that IFC thinks is relevant)
[omit the more funky ones with Normality and Molality]

IonNumberDensity
IonMassDensity
IonChargeDensity

steveraysteveray: I think you are on the right track. We only need to distinguish the quantity kinds that have distinct dimension vectors. Others, like molarity and molality, have to do with moles per volume of solute or moles per volume of final solution, so they have the same dimensionality. We can handle those cases with rdfs:seeAlso. 

Look at our description for quantitykind:Concentration 

So we should list the possibilities by their quantity kinds, as in:
- count per volume
- moles per volume (aka Concentration, AmountOfSubstancePerVolume, AmountOfSubstanceConcentration)
- mass per volume. (aka Density or MassDensity)
- charge per volume

Have to run...

fkleedorfer: right now, all I need is a quntitykind for IonConcentration that uses amount of substance per volume. I'll add just that as `qk:AmountOfSubstanceIonConcentration`

fkleedorfer: done

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
fkleedorfer: done

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
steveraysteveray: I can go along with the skos:broader you suggest.

steveraysteveray (approved):

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
```diff
@@ -11716,6 +11772,19 @@ quantitykind:RotationalStiffness
   rdfs:label "Rotational Stiffness"@en ;
   skos:broader quantitykind:TorquePerAngle .
 
+quantitykind:RotationalVelocity
```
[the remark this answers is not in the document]
fkleedorfer: had not gotten around to it yet

PR #1051, opened 2024-12-03
No longer restrict maven CI to PRs that have been closed

### LINKED ISSUE #1050: Maven CI is not triggered for PRs from forks

fkleedorfer:
Fixes #1050

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -4,7 +4,6 @@
 name: Java CI with Maven
 
 on:
-  push:
   pull_request:
     branches: [ "main" ]
 
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -6,8 +6,6 @@ name: Java CI with Maven
 on:
   push:
   pull_request:
-    types:
-      - closed
     branches: [ "main" ]
 
 jobs:


steveraysteveray (commented):
Does this change mean that the validation is done only when the PR is created (which would not have any changes yet, right?)

Why not have it run with each push? Too much?

If the GitHub defaults were different, I would recommend it runs each time a PR is switched from "draft" to "ready for review", but since the default is "ready for review", people rarely use the "draft" state anyway, so this would not really help.

Or here's an idea: Have the validation run when manually invoked via the Actions tab on GitHub. This avoids too many runs, but allows submitters to validate without any local installation of Maven. We would then request submitters to run it when submitting a PR, and reviewers could also easily run it if it was missed.

fkleedorfer:
As far as I can tell, this runs the CI action every time someone pushes to a PR that is targeted at the main branch

fkleedorfer:
For example, in this PR, the action was executed because of the `pull_request` event, not the `push` event as it would have before.
grafik

steveraysteveray:
So are you saying it runs upon a push and also runs on a pull request?

steveraysteveray (approved):

steveraysteveray:
The action doesn't look like it ran. Also, the latest snapshot release was last week.

fkleedorfer:
True, it does not run on main when the PR gets merged. But I believe it will run on PRs from forks upon each push, possibly after we allow it.

fkleedorfer:
I'll give it another try, adding the push event with the main branch filter...

PR #1049, opened 2024-12-03
adding coefficient of performance to quantitykinds

lazlop:
For HVAC systems we talk about Coefficient of Performance quite a bit, referring to both the capability of systems to heat and cool - seems like a valid addition to qudt!

PR #1054, opened 2024-12-05
Allow manually running maven CI

fkleedorfer:


steveraysteveray (approved):

steveraysteveray:
@fkleedorfer, I wanted to test out the manual validation on PR #1049, but I think it only allows me to choose branches created by code-owners (for security reasons?). That helps us, but not the contributor community. We should discuss what we want, perhaps at the next Board call.

fkleedorfer:
Contributor branches should build now on every push. Maybe they can run our action manually on their branch too

steveraysteveray:
I just tested with my alter-ego, and indeed I can run the validation manually on my fork prior to submitting a PR. It then showed up in our repo asking for our approval to run the validation in qudt-public-repo. So that's pretty useful.

It did not automatically run the validation on the fork following a push, but manual invocation is pretty good. I will update our git practices wiki page when I get the time, to show people how they can validate before submitting a PR.

PR #1057, opened 2024-12-06
add SHACL definition validation in CI

jimkont:
There were some invalid SHACL declarations in the constraint definition. With this PR, I removed them and, additionally, added validation of all the SHACL files using [shacl-shacl](https://www.w3.org/ns/shacl-shacl). The format plugin did not let me keep some comments on the `SHACL-SHACL.ttl` file wrt origin, not sure if you want to add some details in a readme file or if it is ok as is

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,35 @@
                             </validations>
                         </configuration>
                     </execution>
+                    <execution>
+                        <!--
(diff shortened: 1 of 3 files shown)


fkleedorfer:
Love this, btw!

fkleedorfer (approved):

PR #1066, opened 2024-12-06
 Removed unit:CFU

steveraysteveray:


fkleedorfer (approved):
As a sidenote, in some cases (maybe this one, not sure) it might be helpful to add a deprecation explanation a) for our future selves because even the best forget and b) for users stumbling upon the deprecated flag

steveraysteveray:
I'm thinking that when we migrate to QUDT 3.0 we would use that occasion to delete all the deprecated entities. We would have to all agree on that of course.

But your larger point, definitely worth considering. I'm going to go ahead and merge for now.

steveraysteveray:
@fkleedorfer, this PR indicates the check passed, but under the Actions tab it looks like it fails because it was unable to delete the snapshot release. Also, I notice there is no longer a snapshot listed as a Release any more. Is this what you expect?

fkleedorfer:
Well this deleting the release step has never run before. Now that it actually runs, we see it is somehow wrong. Next pr to the rescue. Sorry for the inconvenience

steveraysteveray:
No apologies! The test loop is a slow one here...

PR #1078, opened 2024-12-12
Fix html page links

steveraysteveray:
The latest version and previous version links were being generated incorrectly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 ## [Unreleased]
 
+### Changed
+
+- Removed the vaem:revision triples that were causing retention of v2.1 strings in the URIs
+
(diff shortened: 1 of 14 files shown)


fkleedorfer (approved):
So we do not want vaem:revision triples, is that correct? 

Instead of removing those we could change the publishing algorithm so we don't generate these suffixes?

But I don't have a horse in this race so... Your call

steveraysteveray:
It would be cleaner to change the algorithm, but it lives inside the linkedmodels repository created by @ralphtq, and I try to avoid making changes to it. So, this temporary hack results in the correct links, but also results in the blank line at the top of the page that says: "This version:  ". Removing that line would also require a change to lmdoc. 

Bottom line, all this goes away with our new publishing code when it's ready. At that point, we could decide whether we want to restore the vaem:revision triples for any purpose.

PR #1093, opened 2025-01-09
Create qudt:informativeReference to IEC CDD for units and quantitykinds

### LINKED ISSUE #1092: create qudt:informativeReference to IEC CDD for 1215 units and 211 qk
There are 1215 units with `qudt:iec61360Code` (IRDI).
We should create informative URLs for these units.

There are 2 existing values but the `iec61360/iec61360.nsf` part is wrong:
- https://cdd.iec.ch/cdd/iec61360/iec61360.nsf/Units/0112-2---62720%23UAA105
- https://cdd.iec.ch/cdd/iec61360/iec61360.nsf/Units/0112-2---62720%23UAA924

Example format (using the IRDI from https://github.com/qudt/qudt-public-repo/issues/1091):
```ttl
# from
qudt:iec61360Code "0112/2///62720#UAA609"
# to
qudt:informativeReference "https://cdd.iec.ch/cdd/iec62720/iec62720.nsf/Units/0112-2---62720%23UAA609"^^xsd:anyURI
```
IEC 62720 is about units of measure, and it appears 3 times in the URL (this URL is correct and resolves):
https://cdd.iec.ch/cdd/iec62720/iec62720.nsf/Units/0112-2---62720%23UAA609

I got the property `qudt:informativeReference` from the existing examples.
But maybe it's better to create a dedicated property for such kind of links? (cc @steveraysteveray)

fkleedorfer:
Fixes #1092

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
                   initialize               delete-applicable-units, format-sources, copy-rdf, copy-docs, copy-root-files
                   generate-sources         validate-shacl-files
                   process-sources          replace-version-placeholder
-                  generate-resources
+                  generate-resources       generate-iec-links
                   process-resources
                   compile                  shacl-infer, merge-applicable-units
                   process-classes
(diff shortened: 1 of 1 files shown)


fkleedorfer:
@VladimirAlexiev please review

steveraysteveray (commented):
@fkleedorfer, do I understand correctly that this pom.xml change will change the information references in the src or in the target folder, or both? 
If it changes the src files, this bothers me because we should just change those once and not run this every build.
If it changes the target files but not the src, this bothers me because they will never be consistent.
If it changes just the target files, this bothers me because again the src and target will be inconsistent.

Could you comment?

steveraysteveray:
An alternative would be a query that ADDS an appropriate information reference to the src if it does not exist.

fkleedorfer:
Currently it vhanges only the files in the target folder. Is there a problem with having multiple informative references? If not, we just add the triple.

fkleedorfer:
@steveraysteveray I made the change to add instead of replace the `qudt:informativeReference` links. It works (ie, we have not SHACL rule that prevents it). I prefer adding to replacing anyway - the manually included links are results of human curation that are valuable - in many cases, more helpful than IEC CDD links, as they provide human readable background information

steveraysteveray (approved):

VladimirAlexiev (commented):
@fkleedorfer you should keep the DELETE. It removes only cdd.iec.ch links, because there are currently a few such wrong links (see the issue for details)

fkleedorfer:
Thanks for the hint! I removed the DELETE because it is quite misleading to have some IEC Web links under `src` that are no longer there in `target`. I think the best option would be to forbid IEC Web links in `src` via SHACL and remove those links.

steveraysteveray:
I agree that we should remove bad links, but what if someone manually adds an informative link for a concept that happens to point to an IEC source? Why would we forbid that?

fkleedorfer:
Yes, there is that. We could assume that all current links are bad or can be regenerated and delete them all, but refrain from disallowing them with SHACL

steveraysteveray:
If I understand @VladimirAlexiev correctly, there are only 2 such bad links. So sure. We could delete those. I'm fine with:
1. Deleting any triples with bad links
2. Generating all these new IEC links as informative references.
3. Allowing people in the future to add informative references in the src folder and having them survive.

Down the road, we could possibly test all links during a build, but that would be terribly slow.

fkleedorfer:
> Down the road, we could possibly test all links during a build, but that would be terribly slow.

You read my mind.

It's on my list - but not for the standard build. A release build would be the right kind of setting for that, or some sort of check required for a release build.

PR #1096, opened 2025-01-10
Stop touching sources during build

fkleedorfer:
Prior to this PR, we are filtering `qudt:applicableUnit` triples from the qk file under src. 

As we do not immediately re-format that file, we end up with changes if the build ends prematurely that have to be rolled back before a new build can run sucessfully. That is a random consequence of a bad idea: touching the sources - driving the point home that the idea is bad.

With this PR, we stop touching the sources and instead introduce a SHACL file that only looks at the sources - which will be needed for more checks in the forseeable future, anyway. In the current form, this check ensures that there are no  `qudt:applicableUnit` triples on any quantitykinds.

As a side-product, the generation of `qudt:applicableUnit` triples was extended to instances of `qudt:CurrencyUnit`.


steveraysteveray (approved):
OK. I'm not sure how useful it is to list all the currencies as applicableUnit, but it is strictly speaking correct.

fkleedorfer:
> OK. I'm not sure how useful it is to list all the currencies as applicableUnit, but it is strictly speaking correct.

It was in the currencies file before, so I had to put it back in. Who knows if someone depends on it?

PR #1098, opened 2025-01-21
Refs/heads/main

mobb:
from EDI: 31 Units, 7 quantitykinds, 5 dimensionvectors

### DIFF
--- a/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
+++ b/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
@@ -1390,7 +1390,6 @@ quantitykind:BiodegredationHalfLife
 quantitykind:BiogeochemicalRate
   a qudt:QuantityKind ;
   dcterms:description "the rate at which a specific chemical element or compound is processed or transformed, as amount per unit mass of the relevant material Unreduced, SI unit is mole * kg-1 * s-1" ;
-  qudt:applicableUnit unit:NanoMOL-PER-GM-HR ;
   qudt:hasDimensionVector qkdv:A1E0L0I0M-1H0T-1D0 ;
   qudt:plainTextDescription "the rate at which a specific chemical element or compound is processed or transformed, as amount per unit mass of the relevant material. Unreduced, SI unit is mole * kg-1 * s-1" ;
   qudt:qkdvDenominator qkdv:A0E0L0I0M1H0T1D0 ;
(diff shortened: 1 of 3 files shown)


src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
```diff
@@ -13527,6 +13596,16 @@ quantitykind:TemperatureRelatedVolume
   rdfs:isDefinedBy <http://qudt.org/2.1/vocab/quantitykind> ;
   rdfs:label "temperature-related volume  "@en-US .
 
+quantitykind:TemperatureVariance_NEON
+  a qudt:QuantityKind ;
+  dcterms:description "Variance for NEON temperature data measured in degrees celcius" ;
+  qudt:applicableUnit unit:DEG_C2 ;
```
steveraysteveray, line 13602: Don't assert the applicable unit in the quantity kind file. Instead, add to the units file:
```
unit:DEG_C2 qudt:hasQuantityKind quantitykind:TemperatureVariance_NEON ;
```

steveraysteveray (commented):
See embedded comment.

src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
mobb: Do I need to rerun the formatting workflow after?

mobb: I re-ran the Format workflow, and got: 
```
Token does not have repo scope on EDIorg/qudt-public-repo

Source was already formatted, nothing was touched.
```

steveraysteveray:
Oddly, the one violation has to do with unit:NUM-PER-MilliL having more than one qudt:symbol.
I don't think you even touched that unit.

        qudt:symbol                  "#/mL" , "/mL";

We will need to clean that out, and then you will need to resync your fork. Sorry about that!

steveraysteveray:
I take it back. Looks like you added NUM-PER-MilliL that got merged with the existing one, but with a different symbol. Does that sound right? Your addition is also missing some triples like the scientific notation conversion multiplier. My suggestion is that you remove your added definition.

mobb:
I missed that duplicate. Will fix

fkleedorfer:
Some time ago I made sure every NUM-PER unit has a symbol using the '#' character for consistency. Please use that one as the symbol. the other one can be a qudt:altSymbol.

steveraysteveray (approved):
There are no violations now. There are a few things missing, such as scientific notation conversion multipliers, or any conversion multiplier (e.g. unit:MicroS2-PER-CM2), but because you all have been curating this for so long, and your community is waiting for the new units, I'm going to go ahead and approve it. I trust you will find the time to come back and add them.

PR #1109, opened 2025-01-28
Improve README.md header

fkleedorfer:


steveraysteveray (approved):
The reason for "dataTypes" was to show why the acronym is QUDT and not QUDD. However, it might just raise questions, so I'm fine either way.

Anybody else have an opinion?

fkleedorfer:
What I found really confusing was the '- public repository' at the end.

fkleedorfer:
I tried different ways of emphasizing the QUDT letters, but it does not seem to work in the title.
maybe if we pushed the explanation of the acronym into normal text, such as:

# QUDT

QUDT: ***Q***uantities, ***U***nits, ***D***imensions and Data***T***ypes

...

fkleedorfer:
Not sure what I tried that did not work, now emphasis in the header does work.

PR #1111, opened 2025-01-29
Correct unconventional qudt:symbol strings

### LINKED ISSUE #1110: Recently added units have unconventional symbols
For example, 

```
unit:GM-PER-M2-YR
  qudt:symbol "g/(m<C2><B2><E2><80><B9><E2><80><A6>yr)" ;
  rdfs:label "grams per square meter per year"@en-us ;
  rdfs:label "grams per square metre per year"@en .
```

There are a number of such units. 
I believe that the labels are also nonstandard (containing 'per' multiple times), but I am not sure about that.

Both problems may be fixed in the future by autogenerating such things, but for now, we need to fix this.

fkleedorfer:
Fixes #1110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


steveraysteveray (approved):
Thanks

PR #1114, opened 2025-01-29
Fix/Add some conversion multipliers

fkleedorfer:
In the first pass, this will not build - the explanatory comments are still in there. 
The multipliers need to be checked and integrated or dropped

### DIFF
--- a/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
+++ b/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
@@ -41093,7 +41093,7 @@ unit:REV-PER-MIN
     #    ├─── unit:REV multiplier: 6.28318531
     #    └──┬ unit:MIN^-1 multiplier: 60.0 (multiplier^-1: 0.01666666666666666666666666666666667)
     #       └─── unit:SEC multiplier: 1.0
-  unit:REV-PER-MIN qudt:conversionMultiplier 0.1047197551666666666666666666666667 ;
+  qudt:conversionMultiplier 0.1047197551666666666666666666666667 ;
   qudt:conversionMultiplier 0.01666666666 ;
   qudt:conversionMultiplierSN 1.666666666E-2 ;
   qudt:expression "$rev/min$"^^qudt:LatexString ;
--- a/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
+++ b/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
@@ -4564,16 +4564,10 @@ unit:CI
 unit:CI-PER-KiloGM
   a qudt:Unit ;
   dcterms:description "1,000-fold of the unit curie divided by the SI base unit kilogram" ;
+  qudt:conversionMultiplier 37000000000.0 ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M-1H0T-1D0 ;
   qudt:hasQuantityKind quantitykind:MassicActivity ;
   qudt:hasQuantityKind quantitykind:SpecificActivity ;
-    # MISSING MULTIPLIER: unit:CI-PER-KiloGM - calculated from factors: 37000000000.0
-    #
-    #  ──┬ unit:CI-PER-KiloGM multiplier: [no conversionMultiplier] (correct: 37000000000.0)
-    #    ├─── unit:CI multiplier: 37000000000.0
-    #    └──┬ unit:KiloGM^-1 multiplier: 1.0 (multiplier^-1: 1.0)
-    #       └─── unit:GM multiplier: 0.001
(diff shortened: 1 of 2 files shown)


fkleedorfer:
@steveraysteveray @ralphtq @jhodgesatmb As I said above, this cannot be merged now because the formatting is off (and there are both the old and the new conversion multipliers for the unit in question.   

I did that deliberately so as to make it easier to check the conversion multipliers. I believe the ones added in this PR are correct, but without a second opinion, I would not want to change them

fkleedorfer:
By the way, this PR highlights the need for a precision specification for numbers. I've seen some work on this by the SI people, maybe we can adopt it.

steveraysteveray:
@fkleedorfer, first let me apologize that I let all those conversion multipliers slip through in PR #1098. While the PR was structurally sound, I did not go in and verify the conversion multipliers. Your factoring verifier has proven itself once again, which leads me to ask whether it is practical to include it in this repo's automatic validation?

Meanwhile, I will review this PR.

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -13297,6 +13312,14 @@ unit:HR-PER-YR
   qudt:applicableSystem sou:SI ;
   qudt:applicableSystem sou:USCS ;
   qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1 ;
+    # MISSING MULTIPLIER: unit:HR-PER-YR - calculated from factors: 0.0001140771161305042208532968286561716
```
steveraysteveray, line 13315: As implied by the URI, this is a confusing unit, because there is a fixed value for the number of hours in a year. The description clarifies the intention however. I'm thinking that we should deprecate this unit in favor of a renamed unit, HR_Activity-PER-YR.

Thinking some more, I think it is cleaner to keep the unit the same, but have it refer to one or more quantity kinds, to include quantitykind:TimeRatio_Activity or quantitykind:ActivityTimeRatio, with a skos:broader to quantitykind:TimeRatio.

Thoughts?

steveraysteveray (approved):
Approved, but please see embedded comment for a possible new quantity kind.

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
fkleedorfer: I don't see the problem. How is it different from MilliM-PER-M?

fkleedorfer:
Yes, @steveraysteveray we could include it but it would have to include the QUDTLib code to be checked out and built on the state of the qudt repo in the PR (because the java code is generated based on the RDF). Once that is done, the checks can be run. That would add a few minutes to the build

fkleedorfer:
It may be easier to trigger these checks once a PR is merged, and to have the system make PRs with corrections if something slipped through.

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
steveraysteveray: I think the difference is that MilliM-PER-M references a quantity kind that gives it a context (in this case, a gradient). It is the lack of a context, for two units of the same dimension but different magnitude, that threw me. Maybe it's just the side of the bed I woke up on today.

steveraysteveray:
I'm fine with your last suggestion. Would that be an automatic check?

fkleedorfer:
Yes...

fkleedorfer:
But I am daunted by the cans of worms this opens

steveraysteveray:
Perhaps we can discuss more next week, but I'm looking at this as simply another validation check that happens to use Java instead of SHACL/SPARQL. (I realize that's a bit of an oversimplification because it also involves a separate repository that builds on this repo). 

Looking forward to learning more about the implications.

fkleedorfer:
I think there is a way to include it in the normal build process, which would be preferable. 

If I am not mistaken, this would allow us to do all the checks and generating we want (conversionMultipliers, symbols, labels, check for dead links etc.) then we could reallly strip contributed triples down to the necessary minimum.

Not sure we want to go all the way on it, but we could. Of course this would tie Qudt quite intimately to Qudtlib. Not sure if we would want that either.

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
fkleedorfer: I don't understand the request and whether you want that new QK now or not, but since you approved the PR, I think we can merge (thus avoid collecting more merge conflicts) and add this QK later if we still think it is needed.

steveraysteveray: I want the new QK but I'm fine with waiting for our next Board call to discuss.

PR #1115, opened 2025-01-30
Fix many unit symbols

### LINKED ISSUE #1112: Many qudt:symbol values with more than one divider symbol (slash)
... for example this beauty:
```
unit:FT-HR-PER-GAL_UK
  qudt:symbol "lb/(gal (UK))/(lb/(ft·h))" ;
```

The regex 
```
qudt:symbol "\(?(.+)\)?/\(?.+\)?/.+\)?
``` 
finds 177 of such symbols in the units file.

fkleedorfer:
Fixes #1112 

This will not build as there are additional qudt:symbol triples with comments explaining how they were formed.

Presumably we will find a few bugs, which would be best handled by fixing them in qudtlib and regenerating the changes.

Once the changes are all ok, we can generate a query that deletes the wrong symbols and then add the correct ones, so we don't really have to merge it all manually.


steveraysteveray (approved):
I spot-checked a bunch of these and your results inspire confidence. Even if there is a small number of errors, the new state is clearly better than what we have now, so I'm going to go ahead and approve.

PR #1122, opened 2025-02-10
 Removed all deprecated entities

steveraysteveray:


fkleedorfer:
How do we know we got them all? Could we add a  sh:Info shape that reports them?

steveraysteveray:
I also realized that the sh:targetClass of the validation test is qudt:Concept, but users could very well be referring to qudt entities from their own instances, so the test isn't very useful for external applications.

fkleedorfer:
Hmm.. The COLLECTION_QUDT_USER_TESTS.ttl shapes are not currently included in the 'shacl-validate' execution. When I add them there and activate the shape in question, the execution time is not any longer, and it finds 54 sh:Info validation messages.

fkleedorfer:
Added a commit with these changes.. works fine

steveraysteveray:
@fkleedorfer, I had forgotten I had two variants of the deprecation validation, one for properties and one for objects of ?s ?p ?o. I had erroneously "corrected" the properties one, thinking it was the objects one. Anyway, I have now re-corrected the properties one (qudt:DeprecatedPropertyConstraint), keeping your updated other changes there. I will now test if they both fire as intended. If you'd like to double-check that would be great. Thanks for watching over my shoulder! 4 eyes are better than 2, clearly!

fkleedorfer:
btw, you can run the shacl-validate execution directly via 
```
mvn shacl:validate@shacl-validate
```

steveraysteveray:
If there is no replacement, the query erroneously passes, because the triple 
`        $this dcterms:isReplacedBy ?newq . `
does not match.
If we really wanted to support this case, we would need to use OPTIONAL.

...and yes, I noticed the duplication as well. I need to think about this a bit more.

fkleedorfer:
I made those changes in the latest commit - feel free to revert and force-push if you don't like it

fkleedorfer (approved):

fkleedorfer:
... and it really looks like there are no deprecation warnings/infos so I think this is ready to merge!

PR #1129, opened 2025-02-15
Unit factorization

fkleedorfer:
# Overview

The pull request enhances the QUDT public repository's build process by introducing new executions for managing source modifications using the seq maven plugin. This aims to automate and streamline the inference and formatting of RDF data, ensuring source files are up-to-date and correctly formatted before the build proceeds.

# Detailed Changes  

-  Added seq maven plugin executions, including infer-and-format, infer-factorUnits, and infer-scalingOf, to handle RDF inference and formatting.
-   Updated the shacl maven plugin to version 1.0.3 for the latest features and fixes.
-   Modified the rdfio maven plugin with new executions for merging inferred triples into source and target files.
-   Updated the spotless maven plugin to ensure consistent formatting of RDF and Markdown files.
-   Added a 'fix' profile to run source modifications early in the build, keeping sources updated.
-   Updated the comment block in the POM to reflect new executions and their phases for clarity.

# New Additions  

-  Added a new quantity kind for Energy to the QUDT ontology, enhancing its coverage of energy-related 
  units and properties.

# Effect:

Adds over 18000 triples to the units file in `src`.

# How to Test

1. Revert the last commit ([277ebcd](https://github.com/qudt/qudt-public-repo/pull/1129/commits/277ebcd0416ac83d8e62ce19cd64b1d53b4def22)) 
2. Do `mvn install`, see it fail
3. Do `mvn -Pfix install`, see it succeed and add a lot of triples to the units file

--- 

## Still to do:

* [x] extend the schema to account for the new properties
* [ ] (WON'T FIX) currency units are not present during generation of factor units, therefore they are added as 'generated base units'. That must be fixed
* [x] apply the automated source changes


fkleedorfer:
Lots to discuss here, I'm sure.

steveraysteveray:
@fkleedorfer, much as in the past, is there a way to allow contributors to push their new units and then run an Action (either manually or automatically - we could discuss) on the web to augment their branch with factors, scaling triples and serialization? They could then pull if they want to continue editing, submit a PR if ready for review and merge (all without local use of Maven).

fkleedorfer:
Absolutely. We just change the maven command in the 'Format Sources...' action to `mvn -Pfix install` (and maybe rename the action to reflect that change)

fkleedorfer:
... line 49 in `.github/workflows/format.yml`, if you want to do it

BUILDING.md
```diff
@@ -0,0 +1,163 @@
...
+- **Process Sources**: Checks that all `.ttl` files in `src/` are formatted and valid against SHACL shapes (e.g., `QUDT_SRC_QA_TESTS.ttl`).
+- **Compile**: Infers triples like `qudt:applicableUnit` and merges them into `target/dist/` files.
+- **Process Resources**: Adds links to IEC standards e.g., `qudt:informativeReference` for units.
+- **Test**: Validates the final output in `target/dist/` against stricter SHACL tests.
+- **Install** (with `-Pzip`): Optionally creates the ZIP.
+The special commands above like `mvn seq:run@infer-factorUnits` let you run inference and merging steps manually, updating the source files directly instead of waiting for the full build.
+
+## Troubleshooting
+
+- **TL;DR**: If you don't know or don't want to dig deep, run:
```
steveraysteveray, line 121: What does TL;DR mean?

steveraysteveray (commented):
The BUILDING.md is a bit difficult to follow, but it is nice and thorough. I'm hoping we can communicate with contributers via the README (with links to the Wiki) by saying:
1. If you have Maven installed, normally just do these steps
   ....
3. If you do not have Maven installed, normally just do these steps
   ....

Installing Maven gives you much more flexibility, etc.

steveraysteveray (approved):

fkleedorfer:
I made the Building file because it is easier to keep it up to date with the repo. You adapt it when you change something, not when it gets merged (and then you forget). Also, the readme can point to the building file and then the repo is more self contained

fkleedorfer:
But we don't have to have it.

stevepublic:
@fkleedorfer, I'm here logged in with my test "public" GitHub id, and find that I cannot run any of the Actions on my test branch (called test-branch). I see them, but there is no button to run them.
I will try digging a bit to see what kind of permissions need to be set.

fkleedorfer:
I think the actions are only available for PRs to main

steveraysteveray:
It seems so, but at present only the automated ones, not workflow_dispatch: ones.
There appears to be a quite complex and dangerous way to open up workflows to the public to manually dispatch, but we would need to proceed very carefully.

BUILDING.md
fkleedorfer: "Too long; didn't read". Often used to mark the abstract of a blog post.

steveraysteveray:
Just to close the loop here for the record - the workflows can indeed be invoked by members of the public on their fork of the repo. I had been having trouble because of the way I was testing.

fkleedorfer:
For units like BEATS-PER-MIN, where we don't want to introduce a basE unIt BEATS, a consistent way of dealing with it would be to deperecate it in favor of NUM--PER-MIN and exclude it from the factorization. Giving it factors that don't coincide with the label opens another can of worms, because we then cannot validate factorizations

fkleedorfer:
Now the BUILDING.md file needs an update  (AI regenerate)...

PR #1147, opened 2025-03-10
1146 begin migrating currencies to the main units graph

steveraysteveray:


fkleedorfer (approved):

steveraysteveray:
A suggestion has been made by @ralphtq that we use a "class preamble" for all currencies, making each URI the concatenation of "CCY_" and the international currency code. e.g. unit:CCY_USD for U.S. Dollar, and unit:CCY_CUP for Cuban Peso. This avoids current and future collisions between currency URIs and any other three-letter unit URIs. I'm inclined to move forward on this, pending any strong objections.

fkleedorfer:
Fine with me

PR #1148, opened 2025-03-11
Enable filtering of SHACL validation results

fkleedorfer:
By default, results of severity Info are filtered out. 

This can be configured at the global level with the maven property `shacl.severity.log`. e.g
```
mvn -Dshacl.severity.log=Info install
```

Every <validation> block can also be configured with <logSeverity>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,6 +10,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <qudtPrevReleaseYear>TODO-set-property-qudtPrevReleaseYear</qudtPrevReleaseYear>
         <qudtPrevReleaseMonth>TODO-set-property-qudtPrevReleaseMonth</qudtPrevReleaseMonth>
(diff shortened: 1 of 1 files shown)


steveraysteveray:
@fkleedorfer, quick question, does 
`mvn -Dshacl.severity.log=Info install` 
inhibit the sh:Info but report the others, or does that command mean "include sh:Info and higher"?

steveraysteveray:
Perhaps related, what is the interpretation of line 13 in pom.xml:

        <shacl.severity.log>Violation</shacl.severity.log>

steveraysteveray (approved):

steveraysteveray:
After looking at your changes in this PR, I conclude that line 13 mentioned above means "only add Violations to the log" (by default, right?)

fkleedorfer:
That's correct

fkleedorfer:
> @fkleedorfer, quick question, does `mvn -Dshacl.severity.log=Info install` inhibit the sh:Info but report the others, or does that command mean "include sh:Info and higher"?

info and higher

PR #1172, opened 2025-03-27
used qudt:unit vs qudt:Unit, and removed lang tag on plainTextDescription

alanruttenberg:
In my earlier submission inadvertently used qudt:unit as the type vs qudt:Unit. This fixes that.

### DIFF
--- a/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
+++ b/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
@@ -17032,8 +17032,8 @@ unit:J-PER-M2
   rdfs:label "Joule per Square Metre"@en .
 
 unit:J-PER-M2-SEC0pt5-K
-  qudt:conversionMultiplier "1"^^xsd:decimal ;
-  qudt:conversionMultiplierSN "1.0"^^xsd:double ;
+  qudt:conversionMultiplier 1.0 ;
+  qudt:conversionMultiplierSN 1.0E0 ;
   qudt:hasDimensionVector qkdv:A0E-2L2I0M1H-1T-2.5D0 ;
   qudt:hasQuantityKind quantitykind:ThermalInertia ;
   qudt:plainTextDescription "A unit of thermal inertia measured as Joules per meter squared per square root Seconds per degree Kelvin " ;
@@ -30475,7 +30475,7 @@ unit:MicroW
 
 unit:MicroW-PER-CentiM2-MicroM-SR
   a qudt:Unit ;
-  qudt:conversionMultiplier "10000"^^xsd:decimal ;
+  qudt:conversionMultiplier 10000.0 ;
   qudt:conversionMultiplierSN 1.0E4 ;
   qudt:hasDimensionVector qkdv:A0E0L-1I0M1H0T-3D0 ;
   qudt:hasQuantityKind quantitykind:SpectralRadiance ;
@@ -47906,7 +47906,7 @@ unit:W-PER-M2-M-SR
 
 unit:W-PER-M2-MicroM
   a qudt:Unit ;
-  qudt:conversionMultiplier "1000000"^^xsd:decimal ;
+  qudt:conversionMultiplier 1000000.0 ;
   qudt:conversionMultiplierSN 1.0E6 ;
   qudt:hasDimensionVector qkdv:A0E0L-4I0M1H0T-3D0 ;
   qudt:hasQuantityKind quantitykind:SpectralEmittance ;
@@ -47920,7 +47920,7 @@ unit:W-PER-M2-MicroM
 
 unit:W-PER-M2-MicroM-SR
   a qudt:Unit ;
-  qudt:conversionMultiplier "1000000"^^xsd:decimal ;
+  qudt:conversionMultiplier 1000000.0 ;
   qudt:conversionMultiplierSN 1.0E6 ;
   qudt:hasDimensionVector qkdv:A0E0L-1I0M1H0T-3D0 ;
   qudt:hasQuantityKind quantitykind:SpectralRadiance ;
--- a/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
+++ b/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
@@ -17033,7 +17033,7 @@ unit:J-PER-M2
 
 unit:J-PER-M2-SEC0pt5-K
   qudt:conversionMultiplier "1"^^xsd:decimal ;
-  qudt:conversionMultiplierSN "1"^^xsd:double ;
+  qudt:conversionMultiplierSN "1.0"^^xsd:double ;
   qudt:hasDimensionVector qkdv:A0E-2L2I0M1H-1T-2.5D0 ;
   qudt:hasQuantityKind quantitykind:ThermalInertia ;
   qudt:plainTextDescription "A unit of thermal inertia measured as Joules per meter squared per square root Seconds per degree Kelvin " ;
@@ -30476,7 +30476,7 @@ unit:MicroW
 unit:MicroW-PER-CentiM2-MicroM-SR
   a qudt:Unit ;
   qudt:conversionMultiplier "10000"^^xsd:decimal ;
-  qudt:conversionMultiplierSN 1E4 ;
+  qudt:conversionMultiplierSN 1.0E4 ;
   qudt:hasDimensionVector qkdv:A0E0L-1I0M1H0T-3D0 ;
   qudt:hasQuantityKind quantitykind:SpectralRadiance ;
   qudt:plainTextDescription "A unit of spectral radiance that is the power radiating from a surface per unit of solid angle per unit of wavelength, measured in units of watts per centimeter squared per micrometer per steradian" ;
@@ -47907,7 +47907,7 @@ unit:W-PER-M2-M-SR
 unit:W-PER-M2-MicroM
   a qudt:Unit ;
   qudt:conversionMultiplier "1000000"^^xsd:decimal ;
-  qudt:conversionMultiplierSN 1E6 ;
+  qudt:conversionMultiplierSN 1.0E6 ;
   qudt:hasDimensionVector qkdv:A0E0L-4I0M1H0T-3D0 ;
   qudt:hasQuantityKind quantitykind:SpectralEmittance ;
   qudt:hasQuantityKind quantitykind:SpectralIrradiance ;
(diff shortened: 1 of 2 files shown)


steveraysteveray:
@alanruttenberg, please note the failed validation check for the PR. 

image
As it indicates if you click on "CI Build", you need to make a change to the CHANGELOG.md file with each PR.

steveraysteveray:
Are you able to execute 'mvn spotless:apply' locally on your machine, or on the website for your fork (via Actions)?

steveraysteveray (changes requested):
There are 3 ocurrences of qudt:conversionMultiplierSM rather than qudt:conversionMultiplierSN

alanruttenberg:
Regarding spotless apply, locally I am still getting "Formatted RDF is not isomorphic with original, which means that formatting changed the data." even though I am now working in a branch. Do you have exact instructions for doing it using the github action? I don't see it in your documentation. As I mentioned earlier, I am not familiar with this corner of GitHub.

alanruttenberg:
BTW, It might be worth adding a note to the submission guides explicitly pointing to the build documentation and what specifically needs to be done for a submission vs a build, adding as well the instructions for what to do if spotless:apply fails.

alanruttenberg:
The check gives errors but I don't know how to interpret them or how to fix them. 
failure

alanruttenberg:
Another thing to add to the submission guides: To check the CI failures and what to do given various kinds of errors. As well as a note that the changelog should always be updated.

steveraysteveray:
Have you looked at [this wiki](https://github.com/qudt/qudt-public-repo/blob/main/BUILDING.md)? It covers some of the items, including the mvn spotless:apply situation. But usability of our guides is key, so if it is too buried, that's good feedback.

steveraysteveray:
The Github actions button is described (buried?) [here](https://github.com/qudt/qudt-public-repo/wiki/GoodGitPractices) as step 10. We probably ought to combine our older "GoodGitPractices" page with our newer "Installing QUDT for Contributors" page...

alanruttenberg:
I have reviewed that. Generally, building does not necessarily entail something to do with submission. For instance the build instructions for a java project say how to build the jar files, so you can use them. If I was testing then I would build (whatever that mvn command is, it's not "install") but installing is not something I would generally do. I'd rather put my jars where I would like them.  The building guide does discuss submission at the end, but it's easy to skip reading all that detail because you aren't planning to "build". 

The section "Fix Sources Early" wasn't clear to me.
>When to use: If you’ve added new units or prefixes and need to update VOCAB_QUDT-UNITS-ALL.ttl before validation.
If you've added new units you have already updated VOCAB_QUDT-UNITS-ALL.ttl. I don't really know what you mean by fixing. If you mean validating after you have made changes then say that. It's also not clear why an install is necessary for submission.  Also, that command, like the other mvn commands, fail on my machine. For validation, the idiom, I believe, is to run mvn test. Using install for this is confusing.

As we read from the top we get to "Special Commands for RDF Tasks", but it says run them when needed. It's easy to stop reading there if the explanation of what they are needed for comes first. Only near the end is it explained that some of it is needed for contributing.  I would split out the contributing part into a separate page and link it to the building instructions. Maybe even repeat just those instructions from the build that are necessary in order to do a pull request. 

On QUDT.org I would add to the resources section a link to the overall contribution process above the link for the unit submission guide, and add the other two guides there as well. Or have one link for contributing and then have links to the individual guides from that document. And mention that the Changelog needs to be updated as part of the PR. In my more general experience it is not common that someone not part of the project would edit the Changelog. 

In addition, the guide points to the local use of mvn, but does not explain how to use the Github actions. If it's not uncommon occurrence that there's a problem with the mvn instructions and it is desired that contributors use the actions instead, that should be documented.

One thing that isn't clear to me, and which made me nervous, is what happens when you run these tasks. Are they further modifying the RDF source files, i.e. adding something to the PR other than what you have directly edited? Is that what formatting would do, as is suggested in the error message.  That sort of thing worries me. I don't want to have to sort out, in a diff, what changes I made from what changes have been automatically done from the source. If that's what you are doing, please reconsider.

I understand the rationale for the steps that you want submitters to take, but understand that the level of work being asked for is outside the norm in terms of how complicated it is. There's a balance to be had in what work a contributor needs to do vs what work a maintainer has to do, and I worry that the current process doesn't get that balance right. The consequences of making it too complicated is that people are turned off from submitting and you get what may have happened with Allotrope. Lots of additions not fed back to the project.

Finally, a reminder: I am not writing these notes to complain to you. I'm writing them as what I hope are constructive suggestions that would contribute to the further success of the project. And I want to thank you again for your engagement. I would have given up a while ago had you not been so responsive!

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -17033,11 +17033,12 @@ unit:J-PER-M2
 
 unit:J-PER-M2-SEC0pt5-K
   qudt:conversionMultiplier "1"^^xsd:decimal ;
-  qudt:conversionMultiplierSM "1"^^xsd:double ;
+  qudt:conversionMultiplierSN "1"^^xsd:double ;
```
steveraysteveray, line 17036: This should be 1.0E0

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -30473,9 +30474,9 @@ unit:MicroW
   rdfs:label "Microwatt"@en .
 
 unit:MicroW-PER-CentiM2-MicroM-SR
-  a qudt:unit ;
+  a qudt:Unit ;
   qudt:conversionMultiplier "10000"^^xsd:decimal ;
-  qudt:conversionMultiplierSM 1E4 ;
+  qudt:conversionMultiplierSN 1E4 ;
```
steveraysteveray, line 30479: This should be 1.0E4

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -47904,26 +47905,28 @@ unit:W-PER-M2-M-SR
   rdfs:label "Watts per square metre metre steradian"@en .
 
 unit:W-PER-M2-MicroM
-  a qudt:unit ;
+  a qudt:Unit ;
   qudt:conversionMultiplier "1000000"^^xsd:decimal ;
-  qudt:conversionMultiplierSM 1E6 ;
+  qudt:conversionMultiplierSN 1E6 ;
```
steveraysteveray, line 47910: This should be 1.0E6

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -47904,26 +47905,28 @@ unit:W-PER-M2-M-SR
...
   rdfs:seeAlso quantitykind:Emittance ;
+  qudt:applicableSystem sou:SI ;
   rdfs:seeAlso quantitykind:Irradiance .
 
 unit:W-PER-M2-MicroM-SR
-  a qudt:unit ;
+  a qudt:Unit ;
   qudt:conversionMultiplier "1000000"^^xsd:decimal ;
-  qudt:conversionMultiplierSM 1E6 ;
+  qudt:conversionMultiplierSN 1E6 ;
```
steveraysteveray, line 47924: This should be 1.0E6

alanruttenberg:
Forcing the addition of the exponent is an example where you can simplify with no loss. The lexical space for xsd:double is [documented](https://www.w3.org/TR/xmlschema-2/#double) as "double values have a lexical representation consisting of a mantissa followed, **optionally**, by the character "E" or "e", followed by an exponent" [my emphasis].

alanruttenberg:
if you remember, ping me in the fall, when I'll have more time, and perhaps I can write some additional validation queries to catch the sort of typos you are manually identifying.

steveraysteveray:
I'll do that. Thanks!

steveraysteveray:
@alanruttenberg, to your earlier question about how to interpret a validation failure, I just wanted to make sure you are aware of the validation report that is built, found at target/validation/validationReport.ttl.

Anyway, aside from the need to serialize your edited units code (good ol' mvn spotless:apply), I fetched your branch, did that, ran the build, and the validation report says:
```

[ rdf:type     sh:ValidationReport;
  sh:conforms  false;
  sh:result    [ rdf:type                      sh:ValidationResult;
                 sh:focusNode                  unit:MicroW-PER-CentiM2-MicroM-SR;
                 sh:resultMessage              "unit:MicroW-PER-CentiM2-MicroM-SR qudt:conversionMultiplier is 10000, which does not match the value of qudt:conversionMultiplierSN, 1.0E4, converted to decimal notation, 10000.0 .  ";
                 sh:resultSeverity             sh:Violation;
                 sh:sourceConstraint           _:b0;
                 sh:sourceConstraintComponent  sh:SPARQLConstraintComponent;
                 sh:sourceShape                qudt:conversionMultiplierSnShape;
                 sh:value                      unit:MicroW-PER-CentiM2-MicroM-SR
               ];
  sh:result    [ rdf:type                      sh:ValidationResult;
                 sh:focusNode                  unit:W-PER-M2-MicroM;
                 sh:resultMessage              "unit:W-PER-M2-MicroM qudt:conversionMultiplier is 1000000, which does not match the value of qudt:conversionMultiplierSN, 1.0E6, converted to decimal notation, 1000000.0 .  ";
                 sh:resultSeverity             sh:Violation;
                 sh:sourceConstraint           _:b0;
                 sh:sourceConstraintComponent  sh:SPARQLConstraintComponent;
                 sh:sourceShape                qudt:conversionMultiplierSnShape;
                 sh:value                      unit:W-PER-M2-MicroM
               ];
  sh:result    [ rdf:type                      sh:ValidationResult;
                 sh:focusNode                  unit:W-PER-M2-MicroM-SR;
                 sh:resultMessage              "unit:W-PER-M2-MicroM-SR qudt:conversionMultiplier is 1000000, which does not match the value of qudt:conversionMultiplierSN, 1.0E6, converted to decimal notation, 1000000.0 .  ";
                 sh:resultSeverity             sh:Violation;
                 sh:sourceConstraint           _:b0;
                 sh:sourceConstraintComponent  sh:SPARQLConstraintComponent;
                 sh:sourceShape                qudt:conversionMultiplierSnShape;
                 sh:value                      unit:W-PER-M2-MicroM-SR
               ]
] .
```
i.e. the conversion multipliers are being specified like this:
```
  qudt:conversionMultiplier "1000000"^^xsd:decimal ;
```

We can debate the robustness of our validation test, qudt:conversionMultiplierSnShape, which uses string parsing to compare the conversion multiplier and the scientific notation version, but in the meantime, could you just turn those three into things like
```
  qudt:conversionMultiplier 1000000.0 ;
```
... like all our other conversion multipliers?

And while you are at it, how about changing
```
unit:J-PER-M2-SEC0pt5-K
  qudt:conversionMultiplier "1"^^xsd:decimal ;
  qudt:conversionMultiplierSN "1.0"^^xsd:double ;
```
to
```
unit:J-PER-M2-SEC0pt5-K
  qudt:conversionMultiplier 1.0 ;
  qudt:conversionMultiplierSN 1.0E0 ;
```

We can have a debate about all this afterwards!

alanruttenberg:
Doc says:
qudt:conversionMultiplier . The multiplier to convert quantities using this unit to quantities using the SI unit for this quantity kind, expressed as an xsd:decimal

perhaps change "an xsd:decimal" to "a decimal number with a decimal point and at least one following digit"

steveraysteveray:
Agreed. Either that or we revisit the validating code.

Meanwhile, do you need help with resolving the failing check below? It's just the serialization.

alanruttenberg:
The instructions say to use the action "Format sources and commit to branch" the action intended is labeled, I believe "Fix sources and commit to branch"

steveraysteveray (approved):

PR #1187, opened 2025-05-12
Added Millibar Absolute

Toby-Broom:
Hi All, 

Added a a new unit for our projects, let me know if there is a mistake.

T

### DIFF
--- a/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
+++ b/src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
@@ -30721,6 +30721,22 @@ unit:MilliBAR
   rdfs:isDefinedBy <http://qudt.org/$$QUDT_VERSION$$/vocab/unit> ;
   rdfs:label "Millibar"@en .
 
+unit:MilliBAR-A
+    a qudt:Unit ;
+    dcterms:description "The same as a Millibar, however the Absolute is added when there is a mixture of gauge and absolute measurements to ensure there is a clear distinction"^^qudt:LatexString ;
+    qudt:applicableSystem sou:CGS ;
+    qudt:conversionMultiplier 100.0 ;
+    qudt:conversionMultiplierSN 1.0E2 ;
+    qudt:exactMatch unit:HectoPA ;
+    qudt:hasDimensionVector qkdv:A0E0L-1I0M1H0T-2D0 ;
+    qudt:hasQuantityKind quantitykind:ForcePerArea ;
+    qudt:hasQuantityKind quantitykind:VaporPressure ;
+    qudt:iec61360Code "0112/2///62720#UAA810" ;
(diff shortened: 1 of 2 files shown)


src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -30721,6 +30721,22 @@ unit:MilliBAR
   rdfs:isDefinedBy <http://qudt.org/$$QUDT_VERSION$$/vocab/unit> ;
   rdfs:label "Millibar"@en .
 
+unit:MilliBAR-A
```
steveraysteveray, line 30724: At a minumum, the URI would have to be unit:MilliBAR_A, because the hyphen means multiplication.

steveraysteveray (changes requested):

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
Toby-Broom: Done, I noticed mmHg does not have _

steveraysteveray: The symbol is mmHg, but the URI is unit:MilliM_HG as per our naming rules.

src/main/rdf/vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl
```diff
@@ -30721,6 +30721,22 @@ unit:MilliBAR
   rdfs:isDefinedBy <http://qudt.org/$$QUDT_VERSION$$/vocab/unit> ;
   rdfs:label "Millibar"@en .
 
+unit:MilliBAR_A
+    a qudt:Unit ;
+    dcterms:description "The same as a Millibar, however the Absolute is added when there is a mixture of gauge and absolute measurements to ensure there is a clear distinction"^^qudt:LatexString ;
+    qudt:applicableSystem sou:CGS ;
+    qudt:conversionMultiplier 100.0 ;
+    qudt:conversionMultiplierSN 1.0E2 ;
+    qudt:exactMatch unit:HectoPA ;
```
steveraysteveray, line 30730: To avoid confusion for those examining unit:HectoPA, I suggest we change this triple to
```
qudt:exactMatch unit:MilliBAR ;
```
Then, in the definition of unit:MilliBAR we must also have
```
qudt:exactMatch unit:MilliBAR_A ;
```
...to satisfy our validation constraint (qudt:ExactMatchGoesBothWaysConstraint).

steveraysteveray (changes requested):
After more thought, please see additional request.

steveraysteveray (changes requested):
OK, the Units file looks good now. However, your update to CHANGELOG.md is out of sync with main and needs to be fixed. We are already one minor release later than your version.

PR #1189, opened 2025-05-12
Srr allinone

steveraysteveray:
@fkleedorfer, this PR and branch begin with your initial framing of the pom.xml, and add two "stub" files to serve as the ontology declarations for the OWL and SHACL collections. The idea would be to remove the ontology declarations from all the schema and vocab files and union them with the relevant stub files. I would appreciate your guidance on the order in which to do all that, and also how. (Of course, if it is easier for you to just do it, I would not object!)


steveraysteveray:
I don't want to change any of the files in src for any of this. What I want is to have a single file for users to load that contains all the schema and vocab files they need (actually one such file for SHACL users and one such file for OWL users). 

By convention, there is a one-to-one correspondence between a file and a named graph. So for one of these distribution files, there should be just one ontology declaration statement, and it would contain a reference to the graph metadata statement (vaem:GMD...) also in that file. 

I had not quite finished, but I think the vaem:GMD... statement should in turn contain references to all the other vaem:GMD statements that exist within each of the other graphs that have been combined into this distribution file. We could manually list those in the stub files I created, but it's even better if we can auto-generate them with a query.

Further, I think it makes sense to not change any of the rdfs:isDefinedBy triples, so that users can find out where the "normative" definitions reside.

Happy to discuss this design further of course.

fkleedorfer:
In case of SHACL-SPARQL, the validation/inference relies on the prefix declarations on the ontology entity. That's a problem we'd need to solve somehow. 

Easiest way would be to remove no triples from the union, and instead add another ontology entity. If linked to the ones already in the union with some property, as you suggest, that denotes inclusion, one could always programmatically determine which ontology entity is the one representing the union.

steveraysteveray:
Good point, but some tools (like TopBraid) get upset when they encounter more than one ontology declaration in a file. It doesn't fail, but it does go against convention. Let me think about this...

steveraysteveray:
OK, I've given this some thought, and have the following proposal as a starting point for conversation:
1. Move all sh:declare prefix statements into a single named graph, but not as part of that graph's ontology declaration. It would still need an ontology declaration statement, however, to be found by applications following owl:imports statements.
2. Change all occurrences of sh:prefixes throughout our files to point to this single named graph
3. Add this graph to the imports closure defined in src/main/rdf/schema/SCHEMA-FACADE_QUDT.ttl (this is to keep the modular graph version working properly)
4. When creating the SHACL all-in-one graph, do the following:

a. Merge all the SHACL schema, vocab, src/main/rdf/validation/COLLECTION_QUDT_USER_TESTS.ttl, into the destination SHACL graph
b. Remove all the ontology declaration statements from the destination graph
c. Merge in the src/build/srcgen/collections/SHACL-Collection.ttl graph
d. Serialize and clean up

For the OWL version:
a. Skip the single sh:declare graph defined in step 1, since no SHACL rules are relevant.
b. Merge all the OWL schema, vocab files into the destination OWL graph
c. Remove all the ontology declaration statements from the destination graph
d. Merge in the src/build/srcgen/collections/OWL-Collection.ttl graph
e. Serialize and clean up

I think this approach would retain the SHACL functionality needed within both the allinone graph, and with the separate graphs using owl:imports. I might have missed something though, since this is in my head!

What I don't have, however, is the Maven expertise to execute all this, and look to @fkleedorfer for those skills.

fkleedorfer:
That should work. [SHACL does not require sh:prefixes to point to an owl:Ontology](https://www.w3.org/TR/shacl/#sparql-prefixes):
> The recommended [subject](https://www.w3.org/TR/shacl/#dfn-subject) for values of sh:declare is the IRI of the named graph containing the shapes that use the prefixes. These IRIs are often declared as an instance of owl:Ontology, but this is not required. 
Our SHACL stuff needs a bit of a cleanup anyway. removing

However, in the draft PR I'm working on I've got some changes to the current SHACL stuff, as well as a change to the maven-rdfio-plugin (which we would need here too), I suggest we do that first, rebase this PR afterwards and then complete the work as you suggest here.

fkleedorfer:
@steveraysteveray I think we are done here

fkleedorfer:
Wait.. I did not take into account the stubs that you made. Feel free to tweak the solution to include those in the all-in-one file. Not sure how you wanted it.

steveraysteveray:
I just noticed this line in the pom.xml sparqlUpdate:
`
FILTER(?ontology != <http://qudt.org/$$QUDT_VERSION$$/qudt/allInOne>)
`
...but somehow it didn't seem to avoid deleting the ontology declaration (at least when I ran the build.)

steveraysteveray:
@fkleedorfer, I think it is working as intended! I have test-loaded them in TopBraid, and the OWL one in Protege, which loads without error. In TopBraid, the SHACL validation (just the user-tests, as intended) runs correctly and identifies all the deprecated quantity kinds (at least the ones referred to by a unit). If you are comfortable with what I changed in the pom.xml, I will merge.

PR #1211, opened 2025-06-05
My theory is that spotless has trouble with in-line comments, at least within embedded scripts

steveraysteveray:
Moved the few in-line comments to their own line.


steveraysteveray:
@fkleedorfer, this doesn't "fix" the problem, it just avoids the problem!

fkleedorfer:
We should probably just exclude pom.xml from formatting

steveraysteveray:
I was trying to figure out whether pom.xml gets serialized, but because I'm not really familiar with Maven, I couldn't figure it out. But yes, if you are willing and able to make that change, that would be another way to avoid the problem.

PR #1221, opened 2025-06-30
Fix wrong scalings of unit:GM

fkleedorfer:
Prefix-scalings of GM (e.g. `unit:DecaGM`) are no longer `qudt:scalingOf unit:GM` but `qudt:scalingOf unit:KiloGM`. The former
  could be seen as implying that its conversion factor (e.g. `0.01` for `unit:DecaGM`) would convert toward `unit:GM` instead of `unit:KiloGM`.


steveraysteveray:
Many of your queries use ?this. TopBraid expects $this and chokes on ?this. Can $this be used in the Maven environment?

fkleedorfer:
?var and $var are equivalent according to the standard. If you prefer $ because of TB, we can change that.

steveraysteveray:
It seems TopBraid is not so flexible...

steveraysteveray:
Just for ?this. All the other variables can have ?

fkleedorfer:
If you want to push that in a commit, go ahead!

fkleedorfer:
I don't expect any problems with that change

steveraysteveray:
OK, it turns out it wasn't a problem with $this vs ?this (although I did change them all). It was TopBraid flagging the fact that the sub-SELECT statement was not returning $this (or ?this) in the qudt:ScalingOfAndFactorUnitsPresent validation query, which is apparently a SPARQL violation. I included $this, and also had to add the GROUP BY $this on line 527. Now Maven and TopBraid are both happy.

My only question to you is whether you think the query still works!

fkleedorfer:
We should have unit tests for each shape. No way to tell otherwise - in the long run.

fkleedorfer:
> was not returning $this (or ?this) in the qudt:ScalingOfAndFactorUnitsPresent validation query, which is apparently a SPARQL violation

I was not aware of a sparql violation in the execution of the shacl validation - The new version looks ok too and might be faster.

PR #1223, opened 2025-07-02
Improve consistency checks and fix identified issues

fkleedorfer:
New consistency checks:
 - prevent `qudt:hasDimensionVector` to differ from the combined dimension vectors of the `qudt:hasFactorUnit`s if present
 - prevent `qudt:hasDimensionVector` to differ from the `qudt:scalingOf` if present
 - prevent mixing `qudt:scalingOf` and `qudt:hasFactorUnit`
 - prevent missing `qudt:deprecated`
 - if a dimension vector is missing and can be derived, there is an (additional) violation message containing the derived dimension vector
 - prevent a scaled unit (eg unit:MegaXYZ) from not having a base unit (eg. unit:XYZ)

Inferences have been sped up by an order of magnitude


fkleedorfer:
work continued on #1221

PR #1233, opened 2025-07-14
1232 clean up referrals to deprecated concepts

steveraysteveray:

PR #1234, opened 2025-07-15
Replace deprecated entities

### LINKED ISSUE #1232: Clean up referrals to deprecated concepts

fkleedorfer:
Fixes #1232 

Replaces PR #1233 

With this PR, you can do
```
mvn -Pfix install 
```
right off the bat. If there are things to fix in src, those fixes will be applied.

It is conceivable that a subsequent `mvn install` will detect new errors - but I don't expect that to happen.

Still, right now you can only be sure there are no more errors if you run `mvn install` and it succeeds.


steveraysteveray (approved):
At some point I'd like a brief walk-through of the execution flow of pom.xml

fkleedorfer:
With the utmost pleasure!

PR #1252, opened 2025-07-29
Srr cleanup facade

steveraysteveray:
Removed a couple of import statements from facade.
Removed two no-longer used files.
@fkleedorfer, the all-in-one file you had created was replaced by the two files found in srcgen/collections.


fkleedorfer:
Looks good!

I don't love that the ontology entity files should go into `src/build/srcgen` though.

The name `srcgen` was not great to begin with but now it's a misnomer. The point of the folder (whatever its name, if we choose to rename it) is to have a single entry point for generating shacl shapes/rules from a given sparql query (this happens in the groovy script that is executed at the beginning of the build). One folder per query. Nothing else should be in that folder.

Why don't we:
* rename `srcgen` to `sparql2shacl`
* put those ontology files in a different folder with a specific name, so that there will never be a temptation to put something unrelated in there (I don't have a good recommendation for that name though...,`allinone-descriptors` maybe?

steveraysteveray:
I'm fine with moving that folder out of srcgen. I had put them there when guessing at its purpose. I will move that folder up a level directly under build, but as for the name, I was trying to anticipate that there might be other "collections" beyond just the all-in-ones, possibly our new profiles could benefit from custom ontology declarations and metadata as well. Hence the word collections. How about something like `custom-collection-declarations`?

fkleedorfer (approved):

PR #1256, opened 2025-07-31
Srr wikidata link

steveraysteveray:
Defined qudt:wikidataMatch and populated this new property with links to the Wikidata units and quantity kinds (or equivalent).


fkleedorfer (approved):
Looks good, I would prefer the wikidata uri prefix in the defined prefixes of the unit/qk ttl files, though

PR #1265, opened 2025-08-20
Move build validation files to src/build

### LINKED ISSUE #1263: Missing/incorrect imports
When I try to verify the `owl:imports` for all the various graphs (using release 3.1.4), I see there are a few issues:

https://github.com/qudt/qudt-public-repo/blob/5a19ef66a5b8d8c404f469244304afc7d9f83eaa/src/main/rdf/schema/shacl/SCHEMA_QUDT-DATATYPES_NoOWL.ttl#L8-L20

1. This file imports `<http://qudt.org/shacl/functions#>` which does not seem to exist, but is it supposed to be `<http://qudt.org/$$QUDT_VERSION$$/schema/extensions/functions>` instead?

https://github.com/qudt/qudt-public-repo/blob/5a19ef66a5b8d8c404f469244304afc7d9f83eaa/src/main/rdf/schema/SCHEMA-FACADE_QUDT.ttl#L6-L14
2. This file imports `<http://qudt.org/3.1.4/collection/qa/all>` which does not seem to exist, and I'm not sure what it is supposed to be.
3. This file imports `<http://qudt.org/3.1.4/inference/owl-subset>` which also does not seem to exist, and I'm not sure what it is supposed to be either.

I'd be happy to submit a PR to fix these if it could be determined what these imports need to be (or if they need to be removed).

fkleedorfer:
Moved
* qudt-shacl-functions.ttl
* COLLECTION_QUDT_QA_TESTS_ALL.ttl from src/main/rdf/validation
to src/build/validation

Fixes #1263


fkleedorfer:
ad 1: fixed
ad 2: I removed any references to the QA_ALL file from src/main as well.  One was wrong, but the other reference causes the problem you mention. I wrote this in the commit message:
```
Remove reference to QA_ALL from schema facade

In client-oriented graphs we cannot point to build oriented graphs. If the reference to the QA_ALL graph is needed for development purposes, it has to be brought in with a build-oriented import tree under src/build that includes the client-oriented import tree.
```

If you agree, please add a file to `src/build/[whatever]` that lets you use the QA_ALL file in TopBraid right in this PR - it's done from my point of view.

steveraysteveray:
Closed this one due to final tweaks completed in #1267.

PR #1268, opened 2025-08-21
Add Reciprocal units

### LINKED ISSUE #1259: Reciprocal Units
A curious requirement has come up: frequency measured as 'once every so often' (e.g., once every 3 days)

In this case, the quantity value would modify the unit expression's denominator: 1/(3 days). 

I don't know what the concrete example is, but I think a maintenance use case from facility management would be realistic, e.g., "check the validity of the fire extinguishers every 4 months", and we want to put 4 in that box, not 0.25 in a `unit:PER-MO` box, which would be the obvious workaround.

The actual problem here is that it seems logical to convert the 'once every so often' value into a frequency, but that is wrong, because the two don't have the same dimension vector: they are **reciprocal** to each other - a notion QUDT does not support.

However, we could annotate them as e.g. `qudt:hasReciprocalUnit` - a triple that can be generated during the build, as we know the unit factorization of each unit.  It may follow that we also generate `qudt:hasReciprocalQuantityKind` for quantity kinds.

Downstream clients could take that into account and offer to convert between unit:DAY-PER-NUM and unit:NUM-PER-DAY, KiloGM-PER-M2 and M2-PER-KiloGM, etc., allowing for converting without additional client code, between '3'(day-per-num) to 'Hz' or '/sec' , if that's what's needed.

Note: just looking at the dimension vector would work for units that are pure with respect to the dimensions present, but not for units that have components not present in the dimension vector (currencies, NUM, CASES, DEATHS, ...), so the factor analysis would be required.

For the unit request at hand, we could define `unit:SEC-PER-NUM` as the base unit and a number of relevant scalings, `unit:DAY-PER-NUM`, `unit:WK-PER-NUM`, `unit:MO-PER-NUM`, `unit:YR-PER-NUM`.

fkleedorfer:
- Added a pom section for calculating reciprocal units
- Added schema entries for qudt:hasReciprocalUnit
- Moved triples from predefined-factors-and-scalings.ttl into units file
- Added the hasReciprocalUnit triples where appropriate

Fixes #1259


steveraysteveray (changes requested):
Almost there, but a few surprises crept in, such as:
image
image
image


It looks like situations where the factor unit involves a judgement call? Not sure what the remedy is without more thought.

steveraysteveray:
Also, some deprecated units got the new triple added. Probably cleaner to leave those untouched (and references to them).

fkleedorfer:
I forgot to take into account the factor unit scalar. That should take care of the wrong ones.

fkleedorfer:
> Also, some deprecated units got the new triple added. Probably cleaner to leave those untouched (and references to them).

In recent similar changes I always also updated the deprecated ones. Makes everything easier as there are no exceptions. Alao, deprecation, I think, means:  this will disappear, but will work until then.

steveraysteveray (approved):
Everything looks great!
It uncovers ideas for future inferences, such as:
1. Multiple reciprocal units could be inferred as qudt:exactMatch (need to make sure this is always true)
2. Various qudt:exactMatch sets should replicate some external such as wikidata links? Again, would need to think about the implications of that.

PR #1274, opened 2025-08-26
Add shape to flag violations of title case labels

fkleedorfer:


fkleedorfer:
@steveraysteveray Here are the changes that would be made by this logic. Most seem fine. Some are debatable. 

the prefix algorithm is very simple: prefix label + lowercase unit label. This needs to be adapted for multi-word unit labels.

Especially those unit labels that are not suitable for prepending a prefix label, such as "International Table Calorie" -> "Kilointernational table calorie". A way to fix those labels would be to always start with the unit that is suitable for prepending, ie, "Calorie (International Table)" -> "Kilocalorie (international table)". 

https://gist.github.com/fkleedorfer/2891c7b16dea238d466eee7ceb9ff68e

fkleedorfer:
Another problem is that we cannot currently generate labels for derived units if there is one factor that's specialized for a regional locale ("Meter"@en-US), and the others are language matches but not region-specialized. Need to see how that can be done.

fkleedorfer:
@steveraysteveray  I think this is ready for review now.

fkleedorfer:
> @fkleedorfer, I need some help with what to look for. Are you updating the src files, or just correcting things in dist?

Changes to the main units file are the main chunk of it. Youl'll have to compare the src units file with the version in main locally to get a sense of the changes.
Here is my last `target/dist/src-errors/errors.txt` file that lists all the additions and deletions that were automatically made before the last commit (the one that changed 15K lines in the units file): https://gist.github.com/fkleedorfer/8bad2a7a408b62002f87d2c9f37e808a

The 'src/build/inspections' are just preliminary steps toward the queries in the 'src/build/inference/labels' files.

fkleedorfer:
(just rebased onto main)

steveraysteveray:
Did the Turtle prefixes not get added back in to the dist units file? When I run the Maven build, only these prefixes are declared, so all the units are full URIs:

@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix dcterms: <http://purl.org/dc/terms/> .

steveraysteveray:
Unfortunately, I went so far as to delete my local branch title-case, checked out a fresh copy from the remote, and did a mvn install. I'm still getting missing prefix declarations in just the units file. All the others are fine.

[Here](https://chatgpt.com/share/68af6f87-534c-8009-a883-136c69cddef1) is the diagnosis from ChatGPT, for what it's worth.

steveraysteveray:
After more testing, ChatGPT now thinks it has to do with one or more SPARQL CONSTRUCT queries running on just the units graph:

Why units is special (likely causes)

Only units is being “rebuilt” (CONSTRUCT/INSERT) mid-pipeline
SPARQL UPDATEs that rebuild a graph (or copy triples between graphs) typically drop the namespace map. If your pipeline does extra shaping just for units—e.g., merging fragments, normalizing labels, generating factor units—then the resulting in-memory model may have no stored prefixes, so the writer emits a minimal set.
Quick test: Grep your SPARQL update files or RDFIO steps for rules that target units only (graph IRIs or file globs). If units has bespoke UPDATEs and the others don’t, that’s your smoking gun.

fkleedorfer:
Apart from that, I think this is done.

It's a lot of new triples, many of them identical except for the language tag, which is not so nice. I thought about removing all labels that are identical with the default label (no language tag). Clients would look for their language and fall back to the default. However, there is no way with this approach to differentiate between a missing label (say, in Hebrew) and a label that is identical with the default.

To reduce file sizes, we could use a placeholder label for the same-as-default case, such as ""@it or "[DEF]"@it, and explain that that means: get the default label instead. But that's nonstandard and potentially causes lots of downstream effort - I don't like it much.

steveraysteveray:
Your last comment made my day! I had assumed you were seeing what I was seeing. I just tried mvn install on my laptop, and it works fine. So it must be the Maven environment on my desktop. I'll try updating it...

fkleedorfer:
nothing wrong with your maven environment, I think - I did not change mine...

What I did update was the version of the [spotless maven plugin](https://github.com/qudt/qudt-public-repo/pull/1274/commits/25904c06a981d1e9d1ba107a12d36b64249b7628)

It's an intermittent error, so not seeing it does not necessarily mean it's fixed. I don't think I've seen it when not using the `fix` profile though, so chances are that we are on the safe side for making the release.

steveraysteveray:
OK. I'm in a rabbit hole of updating my Java, Maven, etc. environment, so I'm going to try to see it through. Meanwhile, I'll go ahead and approve this PR.

steveraysteveray (approved):

PR #1278, opened 2025-09-04
Autogenerate and unify UCUM codes

fkleedorfer:
Includes `qudt:ucumCode` in the autogeneration routines, and changes the SHACL shapes such that only one value for this property is allowed.

UCUM allows multiple ways to write units, the 'canonical' one is the easiest to generate, so that's the one we use throughout the vocabulary.


steveraysteveray (approved):
I'm convinced. Last chance for @ralphtq and @jhodgesatmb to weigh in!

PR #1286, opened 2025-09-17
Fix description of activity quantities

### LINKED ISSUE #1280: Mistakes in quantitykind:Activity and quantitykind:activityConcentration
**quantitykind:Activity** refers to the rate of nuclear decays (as in, radioactivity), see [VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl#L316]( https://github.com/qudt/qudt-public-repo/blob/88770ad9e093f139b793d9d09d6d03e384b3b95e/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl#L316)

However, two properties refer (wrongly) to the nuclear mass number (often abbreviated as A) instead:

*  [L326](https://github.com/qudt/qudt-public-repo/blob/88770ad9e093f139b793d9d09d6d03e384b3b95e/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl#L326): `qudt:informativeReference "http://en.wikipedia.org/wiki/Mass_number"^^xsd:anyURI ;`,

* [L331](https://github.com/qudt/qudt-public-repo/blob/88770ad9e093f139b793d9d09d6d03e384b3b95e/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl#L331) The first line/sentence in `qudt:latexDefinition` , which reads `$A = Z + N$, where $Z$ is the atomic number and $N$ is the neutron number.` 

**These should be removed or replaced.** 

Additionally, the rest of the latexDefinition reads a bit off. 

Currently: 

> Variation $dN$ of spontaneous number of nuclei $N$ in a particular energy state, 
> in a sample of radionuclide, due to spontaneous nuclear transitions from this state during an infinitesimal time interval, 
> divided by its duration $dt$, thus $A = -\\frac{dN}{dt}$.

Should probably better be something like 

> Variation $dN$ of ~spontaneous~ number of nuclei $N$ in a particular energy state, 
> in a sample of **radionuclides**, due to spontaneous nuclear transitions from this state during an infinitesimal time interval, 
> divided by its duration $dt$, thus **activity** $A = -\\frac{dN}{dt}$.

**quantitykind:activityConcentration** has two properties with incomplete sentences: 

* `dcterms:description "The \"Activity Concentration\", also known as volume activity, and activity density, is ."^^qudt:LatexString ;`: [L358](https://github.com/qudt/qudt-public-repo/blob/88770ad9e093f139b793d9d09d6d03e384b3b95e/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl#L358)
* ` qudt:plainTextDescription "The \"Activity Concentration\", also known as volume activity, and activity density, is ." ;`, see [L363](https://github.com/qudt/qudt-public-repo/blob/88770ad9e093f139b793d9d09d6d03e384b3b95e/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl#L363C1-L363C122)

I believe these are missing something like ` the ratio of the activity and the sample volume`

henrikef:
Fixed/completed the natural language descriptions of  `quantitykind:Activity` and `quantitykind:ActivityConcentration`

Fixes #1280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 - Tweaked some multi-line validation error messages for environments that only display the first line (e.g. TopBraid)
 - Fixed `qudt:ucumCode` of many units, such that they all follow the same pattern (see 'Changed' below), and such that
   no unit has more than one.
+- Fixed descriptions of `quantitykind:Activity` and `quantitykind:ActivityConcentration`.  
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 - Tweaked some multi-line validation error messages for environments that only display the first line (e.g. TopBraid)
 - Fixed `qudt:ucumCode` of many units, such that they all follow the same pattern (see 'Changed' below), and such that
   no unit has more than one.
-- Fixed descriptions of `quantitykind:Activity` and `quantitykind:ActivityConcentration`.
+- Fixed descriptions of `quantitykind:Activity` and `quantitykind:ActivityConcentration` ([Henrike Fleischhack](https://github.com/henrikef)).
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


fkleedorfer:
Thanks @henrikef ! You may want to link to your guthub profile in the changelog the way others have done.

fkleedorfer (approved):

PR #1296, opened 2025-10-08
Add query/update that can fix deprecations in client graph

fkleedorfer:
Not sure this is the right way to incorportat this in the repo.

Anyway, with this PR, we introduce a query and an update that allow for:
1. finding usages of deprecated entities
2. replacing those usages with the `dcterms:isReplacedBy` entity.

Limitations:
- if the deprecated entity occurs in the subject position, it is not detected/fixed
- if there is no `dcterms:isReplacedBy` entity, the entity is left in the graph as-is.

To test:
``` 
mvn rdfio:pipeline@inspect
```
Then look into `target/inspection/deprecation/`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -415,6 +415,25 @@
                                         </files>
                                         <toGraphsPattern>dist:schema:${name}</toGraphsPattern>
                                     </add>
+                                    <sparqlUpdate>
+                                        <sparql>
+                                            INSERT {
+                                                unit:LB-PER-GAL_IMP qudt:exactMatch unit:LB-PER-GAL
+                                            }
+                                            WHERE {}
+                                        </sparql>
+                                    </sparqlUpdate>
+                                    <sparqlQuery>
+                                        <file>src/build/inspection/deprecation/find-use-of-deprecated.rq</file>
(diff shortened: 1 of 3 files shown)


fkleedorfer:
@steveraysteveray  is that what you meant?

fkleedorfer:
`mvn rdfio:pipeline@inspect`

fkleedorfer:
This is only for running the inspections in the qudt build, as a kind of demonstration. If you want to run this on your own data, you'll have to execute the query there.

fkleedorfer:
Let's leave it in as an inspection.

steveraysteveray:
Sounds good. Should we remove things like:
[replace-deprecated-with-replacement.rq](https://github.com/qudt/qudt-public-repo/pull/1296/files#diff-62f9de1b970431d40c2f4345f943099b5f69d0218111c7c62282a68382762538)
and the sparqlUpdate?

                                    <sparqlUpdate>
                                        <sparql>
                                            INSERT {
                                                unit:LB-PER-GAL_IMP qudt:exactMatch unit:LB-PER-GAL
                                            }
                                            WHERE {}
                                        </sparql>
                                    </sparqlUpdate>

steveraysteveray:
Closing so that we can merge #1516 cleanly.

PR #1301, opened 2025-10-14
RateOfChange

pranay2811:
- Added quantityKind:RateOfChange & unit:PERCENT-PER-SEC
- Replaced quantityKind:Frequency to quantityKind:RateOfChange for unit:PERCENT-PER-YR & unit:PERCENT-PER-WK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,11 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 ## [Unreleased]
 
-- Added quantityKind:RateOfChange & unit:PERCENT-PER-SEC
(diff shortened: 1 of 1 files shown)


pranay2811:
@steveraysteveray Thank you for your feedback on my previous PR. I’ve incorporated those changes as per your instructions in this one.

pranay2811:
@steveraysteveray Thank you for pointing out the errors. I will keep in mind that I should run "mvn -Pfix install". Is there documentation for this that I can refer to?

pranay2811:
@steveraysteveray Thank you for sharing this. I will take a look at the documentation and will get back if there is any feedback. Again thanks for the detailed explanation.

PR #1310, opened 2025-10-24
TON_UK-PER-HR

pranay2811:


pranay2811:
@steveraysteveray
On the documentation, I found it straightforward and easy to follow. In GoodGITPractices - step 10, while it mentions that MVN can be run from the website, it doesn’t provide the steps for doing so. This is very minor, but I was able to do it because of your instructions in the last PR I submitted. Even if there’s an external link on how to run MVN tools from the web, that might be helpful. Just a thought. Otherwise, I believe I can follow the documentation. It might take a new and inexperienced user like me a few tries to get the hang of it.

steveraysteveray:
Thanks for the feedback. I'll take another look and see if a few screen shots might help in step 10.

PR #1318, opened 2025-10-28
KiloW-PER-TON_FG

pranay2811:
This PR introduces a new unit **KiloW-PER-TON_FG** with a new quantity kind **CoolingPerformanceRatio**.

This addition supports kilowatt per ton of refrigeration, commonly used across HVAC systems to express the input power required per unit of cooling capacity delivered. It is widely used to evaluate and benchmark the energy performance of chillers, cooling plants, and refrigeration systems.

While QUDT already defines ThermalEfficiency and other dimensionless ratio, those represent output-to-input relationships (efficiency or coefficient of performance). The CoolingPerformanceRatio instead expresses the inverse relationship; _input power per cooling output_. 

Since this ratio is dimensionless but semantically distinct from efficiency, I am proposing a new quantity kind, CoolingPerformanceRatio with skos:broader of DimensionlessRatio. I am open to alternative naming suggestions that align more closely with QUDT’s conventions.

The unit: KiloW-PER-TON_FG is derived as:
1 ton of refrigeration = 3.516853 kW, so 1 kW/ton = 0.2843451361

### DIFF
--- a/.github/workflows/fix.yml
+++ b/.github/workflows/fix.yml
@@ -1,6 +1,8 @@
 name: run `mvn -Pfix install`
+# Dummy invocation; this is a reusable workflow that is called by other workflows.
 on:
-  workflow_dispatch:
+  workflow_call:
+#  workflow_dispatch:
 
 jobs:
   format:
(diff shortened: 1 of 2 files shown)


steveraysteveray:
@pranay2811 - whoops, you added the entry in the CHANGELOG.md to the 3.1.7 release. Hard to notice, because yours is the first PR after that. It should have gone between lines 8 and 10, along with a new header. 

I'll fix it next time I do a PR, no worries.

Issue #130, opened 2020-05-31
FRED.stlouisfed.org (coding Financial-Economic Units)


KevinCuddeback:
The St Louis Federal Reserve sponsors an extensive set of data feeds (792,000) that are (1) poorly indexed (they say they don't have a list of all their feeds) and poorly unitized (seem to be "cut-and-pasted" from the prose title, rather than systematically logged.

I'm proposing for myself (and would like help with)
1) To map all 765,000 data feeds (most US Federal sources, like Census, BLS, and financial regulators)
2) To map the long-prose and short-prose units used in all the feeds
3) To code them so that "everyone" in the QUDT community would have a systematic way of understanding when FRED says "mil. of $" as the units that it is expressed in standard QUDT form in some kind of cross-walk

I have begun at  https://fred.stlouisfed.org/tags/series

with the 167 currently-active feeds  (there are hundreds of others that are infrequent or ancient (cotton mill production in New England in the early 1800s, for example)

https://fred.stlouisfed.org/release?rid=9
--
https://fred.stlouisfed.org/release?rid=10
https://fred.stlouisfed.org/release?rid=11
https://fred.stlouisfed.org/release?rid=13
https://fred.stlouisfed.org/release?rid=14
https://fred.stlouisfed.org/release?rid=15
https://fred.stlouisfed.org/release?rid=17
https://fred.stlouisfed.org/release?rid=18
https://fred.stlouisfed.org/release?rid=19
https://fred.stlouisfed.org/release?rid=20
https://fred.stlouisfed.org/release?rid=21
https://fred.stlouisfed.org/release?rid=22
https://fred.stlouisfed.org/release?rid=25
https://fred.stlouisfed.org/release?rid=27
https://fred.stlouisfed.org/release?rid=46
https://fred.stlouisfed.org/release?rid=47
https://fred.stlouisfed.org/release?rid=49
https://fred.stlouisfed.org/release?rid=50
https://fred.stlouisfed.org/release?rid=51
https://fred.stlouisfed.org/release?rid=52
https://fred.stlouisfed.org/release?rid=53
https://fred.stlouisfed.org/release?rid=54
https://fred.stlouisfed.org/release?rid=61
https://fred.stlouisfed.org/release?rid=63
https://fred.stlouisfed.org/release?rid=70
https://fred.stlouisfed.org/release?rid=71
https://fred.stlouisfed.org/release?rid=72
https://fred.stlouisfed.org/release?rid=86
[...]

PR #1322, opened 2025-10-30
 Added a new class and instances for crypto currency units

steveraysteveray:
A little random, but I thought maybe we should include crypto. Looking for feedback really.

@KevinCuddeback, I'm curious what you think also, re Issue #130.


fkleedorfer:
Should not be a problem to do it but shouldn't we wait until someone asks?

PR #1323, opened 2025-10-30
Srr factor unit order

steveraysteveray:
A handful of units with manually specified factor units were constantly being re-serialized during the build. This PR fixes those to be in the correct sequence (I hope!)


fkleedorfer:
The reordering of blank nodes has become a constant nuisance that has to be fixed upstream in the turtle-formatter. Now that we reprocess files many times it's just more likely to happen. It can be fixed but it's actual work.

PR #1328, opened 2025-11-08
Srr reified factor

steveraysteveray:
This PR adds a type declaration to the bnodes describing the factor units, only in the dist and not in the src. In other words, it does not cause a -Pfix error.


steveraysteveray:
I'm fine with any suggested name changes. @ralphtq, how do you feel about @fkleedorfer's suggestion? I know you had some opinions on this.

fkleedorfer:
Even better name: FactorUnit

steveraysteveray:
This might be a symptom of English vs. German regarding which is the adjective and which is the noun.

steveraysteveray:
Hmm. I was just thinking that we could go simple and just have the class "Factor". We could also rename the relation "hasFactor".
They are only relevant for Units anyway.

PR #1336, opened 2025-11-22
Srr rollback factor unit class

steveraysteveray:
Turns out that when declaring FactorUnit class instances, I had introduced some errors in the build process that only manifest during the -Pfix invocation. This PR rolls back the FactorUnit changes.


fkleedorfer:
Thanks for the colorful description!

steveraysteveray:
Haha!

If you approve it, I'll merge so that if anybody tries a push they won't get clobbered.

fkleedorfer (approved):

PR #1340, opened 2025-12-01
 Added a final update query to type the FactorUnit bnodes

steveraysteveray:
This works, except for the bnode duplication in the Turtle formatter problem, so it is still draft.


steveraysteveray:
Superceded by #1341.

PR #1371, opened 2026-02-20
Srr fix owl schema

steveraysteveray:


fredrikvaws (approved):
With reference to our common review. Approved - but we should review future usage of rdfs:Literal as it might be to broad.

PR #1377, opened 2026-03-06
Adding unit:PicoCI, unit:PicoCI-PER-L, unit:GRAIN_AIR

gtfierro:
Based on suggestions at https://github.com/BrickSchema/Brick/issues/758

Adds:
- unit:PicoCI-PER-L
- unit:GRAIN_AIR (psychrometric mass ratio unit)
- I also added `PicoCI`, but I can remove if you'd rather it not be in there.

I've run the `mvn spotless:apply install` and `mvn -Pfix install` as indicated on the wiki and all tests pass.


gtfierro:
Great suggestions! I agree this is a much more elegant choice. Please see the changes I've made to reflect this; I chose to keep the GRAIN-PER-LB_M "generic" in its definition and to put the HVAC context under the SpecificHumidity quantity kind.

PR #1383, opened 2026-03-17
improve definition of time minute and angle minute

Zack-83:
The definition of minute was mixing angle and time. I rewrote both definitions, hoping to make more clarity.


Zack-83:
Dear F Kleedorfer,

thank you for the hints. I tried running the script, but I got two errors which seem not to depend on me:

grafik

PR #1437, opened 2026-04-21
Propulsion extension

steveraysteveray:
Migrated 31 qks, 3 referring propulsion units, and 1 normal unit reference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 ### Changed
 
+- Migrated 31 quantity kinds that are specific to the field of propulsion to the extensions folder.
+
 ### Deprecated
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


steveraysteveray:
@ralphtq, is your test harness able to explore the imports closure, validation, and orphan entities on this branch?

steveraysteveray:
Even though these new graphs pass our current build validation, the target unit file does not contain any inferred factor units. Here are my current thoughts on handling this:

How should we handle factor unit inferencing?
	1. Don't infer them at all, and don't validate for them? This fails in other environments like TopBraid that will still see the extension graphs and flag missing factor units. Maybe that's OK - it is up to the maintainer of the extension to maintain validity.
	2. Infer them, which would need to be handled for each extension individually - doesn't scale well, but maybe it's worth it?
	3. Expect extension providers to manually declare the factor units? This is unwieldy and error-prone.

steveraysteveray:
Superceded by #1438

PR #1441, opened 2026-04-24
Fix ReferenceDatatype, GlobalPositionVector-DP, ONstate, OFFstate, WDST_WET, WDST_DRY, True, Yes

ektrah:
* Moved `qudt:ReferenceDatatype` back from  
  SCHEMA_QUDT-COORDINATES_NoOWL.ttl to  
  SCHEMA_QUDT-DATATYPES_NoOWL.ttl because it seems unrelated to coordinates and it is needed by  
  VOCAB_QUDT-DATATYPES.ttl
* Moved `datatype:GlobalPositionVector-DP` from  
  VOCAB_QUDT-DATATYPES.ttl to  
  VOCAB_QUDT-COORDINATES.ttl because it is an instance of `qudt:StateSpaceVector` defined in  
  SCHEMA_QUDT-COORDINATES_NoOWL.ttl
* Made `datatype:WDST_WET` and `datatype:WDST_DRY` instances of `qudt:WetDryStateType` instead of `qudt:WetDryState` because `qudt:WetDryState` does not exist
* Deleted `datatype:True` because neither `qudt:BooleanStateType` nor `datatype:False` exists (both `datatype:TRUE` and `datatype:FALSE` exist)
* Deleted `datatype:Yes` because it looks like a duplication of `datatype:YES`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 - Removed some erroneous references from the datatypes schema metadata to the coordinate systems schema under construction
 - Fixed some small errors in qudt:ArrayDataOrder, qudt:MassPropertiesArray, qkdv:A0E1L0I0M-1H0T0D0, and unit:RT
+- Fixed the types of datatype:ONstate, datatype:OFFstate, datatype:WDST_WET, and datatype:WDST_DRY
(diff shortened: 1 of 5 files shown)


steveraysteveray:
I defer to @ralphtq on this PR.

ralphtq (approved):

PR #1443, opened 2026-05-05
Adding unit:HectoHZ

evertoncolling:
Adds `unit:HectoHZ` (Hectohertz), the 100-fold SI prefix scaling of `unit:HZ`, for use cases where frequencies are conventionally reported in hundreds of hertz.

I've run `mvn -Pfix install` and all SHACL validations pass (0 violations).


steveraysteveray (approved):
Looks perfect.

fkleedorfer (approved):

PR #1447, opened 2026-05-07
Adding pressure gradient units

evertoncolling:
Adds three pressure-gradient units: `unit:BAR-PER-M`, `unit:PSI-PER-FT`, and `unit:PSI-PER-M`.

These fill gaps around existing pressure-gradient units such as `unit:KiloPA-PER-M`, `unit:HectoPA-PER-M`, `unit:ATM-PER-M`, and `unit:PSI-PER-IN`. The conversions are derived from existing QUDT pressure and length definitions.

I've run `mvn -Pfix install` and all SHACL validations pass (0 violations). I reverted the unrelated factor-unit ordering that were introduced by Spotless, keeping only the new added units.


steveraysteveray (approved):
Excellent work as usual! In the future you might want to give yourself some credit by adding [Everton Colling](https://github.com/evertoncolling) at the end of your changelog entries.

Also optional is to add any qudt:informativeReference triples you think might be helpful.

Thanks.

evertoncolling:
> Excellent work as usual! In the future you might want to give yourself some credit by adding [Everton Colling](https://github.com/evertoncolling) at the end of your changelog entries.
> 
> Also optional is to add any qudt:informativeReference triples you think might be helpful.
> 
> Thanks.

I see. I’ll keep that in mind in my upcoming PRs!

PR #1458, opened 2026-05-11
Srr reset baselines for OWL derivation comparisons

steveraysteveray:


steveraysteveray:
For context, this PR allows for a big cleanup of the inspection files (what is "expected" and "unexpected") because the new baseline OWL files used for comparison are up to date, unlike the legacy files that were used in the original OWL derivation code. That's why a lot of files got changed this time. In the future, whenever we might update the baseline OWL files used for comparison, there will be far fewer changes.

fkleedorfer (approved):

ralphtq (approved):

PR #1460, opened 2026-05-14
Extended dimension vector

steveraysteveray:
## Extended Dimension Vectors (`edv:`)

This branch introduces the `qudt:ExtendedDimensionVector` system, which extends
the seven SI base dimensions with two additional dimensions suppressed in SI but
physically meaningful:

- **R (angle)** — tracks the radian dimension. Examples: `R=1` for plane angle,
  `R=−1` for torque (τ = dW/dθ), `R=−2` for moment of inertia and radiance.
- **N (count)** — tracks discrete enumerable entities. Examples: `N=1` for
  particle number, nucleon number, turn count, heartbeats.

### Dividing line for N=1

`N=1` when the quantity answers "how many of [this discrete physical entity]?"
`N=0` when the quantity is a dimensionless ratio, quantum-mechanical index, or
information-theoretic unit — even if it is always an integer. For example,
`quantitykind:AtomicNumber` and `quantitykind:MassNumber` carry `N=1` (they
count protons and nucleons respectively), while `quantitykind:ChargeNumber`
carries `N=0` (it is q/e, a ratio). `unit:BIT` is `N=0`: bits are information
units — you do not measure information content by counting discrete objects the
way you count protons. Full rationale and worked examples are in
`src/main/rdf/vocab/extendeddimensionvectors/DESIGN-NOTES.md`.

### Why currency is not a third new dimension

R and N are motivated by dimensional disambiguation: torque and energy share the
same SI DV, as do angular momentum and action, and the new dimensions separate
them. Currency does not play that role — no two quantity kinds in QUDT's physical
vocabulary are ambiguous solely because of a missing currency dimension. Currency
is also fundamentally unlike a physical dimension in that there is no fixed
conversion between currency units; exchange rates are not physical constants.
QUDT tracks currency through a dedicated mechanism: units follow the `CCY_`
naming convention (e.g., `unit:CCY_USD`, `unit:CCY_EUR`) and carry the type
`qudt:CurrencyUnit`, with ISO 4217 numeric codes via `qudt:currencyNumber`. That
type-and-naming approach is the right home for currency; adding a C dimension
would provide no dimensional-analysis benefit while implying a spurious
universality across currencies that do not share a stable ratio.

### What's in this PR

**Foundation (commits `9a9011c8`–`b1719915`)** — rebased from the
`infer-dv-properties` branch:
- IRI-based inference of `qudt:dimensionExponentFor*` properties on `qkdv:`
  instances, eliminating ~2,168 explicit triples
- Deprecation of `qkdv:A0E-2L2I0M1H-1T-2.5D0` →
  `qkdv:A0E-2L2I0M1H-1T-2dot5D0` (standardise `dot` encoding for fractional
  exponents)
- Deprecation of `unit:J-PER-M2-SEC0pt5-K` → `unit:J-PER-M2-SEC0dot5-K`;
  factor units added to both

**Schema (`eaeef46b`):**
- `qudt:ExtendedDimensionVector` class (SHACL node shape with `sh:minCount 1`
  on all 10 dimension properties)
- Properties `qudt:dimensionExponentForAngle`, `qudt:dimensionExponentForCount`,
  `qudt:hasExtendedDimensionVector`
- `qudt:QuantityKind-hasExtendedDimensionVector` (structural, `sh:minCount 1`)
- `qudt:Unit-hasExtendedDimensionVector` (cardinality only, `sh:maxCount 1`)
- Build: `infer-dv-exponents.ru` extended to infer R and N exponents on `edv:`
  instances from the IRI local name

**Vocabulary population (`b5e0ef95`):**
- 260 `edv:` instances in `VOCAB_QUDT-EXTENDED-DIMENSION-VECTORS.ttl`
- `qudt:hasExtendedDimensionVector` added explicitly to all 1,213
  `QuantityKind`s and all 2,912 `Unit`s that carry a `hasDimensionVector`
- 79 QKs and 109 units receive non-R0N0 assignments (angular or count); all
  others default to R0N0
- Two QA shapes added to `COLLECTION_QUDT_QA_TESTS_ALL.ttl` (both
  `sh:Warning`):
  - `qudt:ExtendedDVBroaderChainHeuristic` — fires if a QK's `skos:broader`
    ancestor has non-R0N0 edv but the QK itself has R0N0 (catches missing
    assignments on new angular/count specialisations)
  - `qudt:UnitExtendedDVConflict` — fires for units whose `hasQuantityKind`
    values disagree on R or N (documents the 15 known dimensional-cancellation
    cases listed below)

**Corrections and documentation (`60eb93c7`):**
- `unit:BIT`, `BYTE`, `FRACTION`: removed incorrect `qudt:CountingUnit` type
- `unit:NUM`: edv corrected to `R0N1`; `unit:Z`: edv corrected to `R0N1`
- `quantitykind:AtomicNumber`, `MassNumber`, `NucleonNumber`, `NeutronNumber`:
  edv corrected to `R0N1` (discrete particle counts; `AtomicNumber` and
  `MassNumber` already carry `skos:broader Count`)
- All `dcterms:description`, `qudt:plainTextDescription`, and
  `^^qudt:LatexString` description fields in the unit and QK vocabulary
  reformatted to triple-quoted strings with line breaks near 130 characters;
  `qudt:latexDefinition` fields are left unchanged
- Design rationale in
  `src/main/rdf/vocab/extendeddimensionvectors/DESIGN-NOTES.md`

### Known ambiguities (SHACL warnings, not errors)

The following units serve QKs that share the same SI DV but disagree on R or N.
They carry `R0N0` as a conservative default. `qudt:UnitExtendedDVConflict` flags
them as `sh:Warning` on every build; results are written to
`target/validation/validationReport.ttl`.

| Unit | Conflict |
|---|---|
| `unit:J-SEC`, `unit:ERG-SEC`, etc. | `AngularMomentum` (R=−1) vs `Action` (R=0) |
| `unit:PER-M` and scaled variants | `AngularWavenumber` (R=1) vs `InverseLength` (R=0) |
| `unit:COUNT`, `unit:UNITLESS`, `unit:NUM` | count QKs (N=1) vs dimensionless QKs (N=0) |
| `unit:FT-LB_F` | `Torque` (R=−1) vs `Energy` (R=0) |
| `unit:NUM-PER-SEC`, `NUM-PER-YR`, `SAMPLE-PER-SEC` | `CountRate` (N=1) vs `Frequency` (N=0) |

Resolution would require introducing dedicated angular or count unit variants and
is deferred to a follow-on PR.

### Note on `unit:CYC`

In physics, frequency is universally expressed in hertz (Hz = cycles per
second), where "cycle" means one complete oscillation — a dimensionless count,
not an angular displacement. Angular frequency ω is a distinct quantity measured
in rad/s; the relationship ω = 2πf makes explicit that the radian factor is
supplied by the conversion, not by the unit of frequency itself. Treating
`unit:CYC` as `N=1` (dimensionless count, conversionMultiplier 1.0) keeps
`unit:CYC-PER-SEC` exactly equal to `unit:HZ`, consistent with how practising
physicists and engineers use these units. The angular role is covered by
`unit:REV` (conversionMultiplier 2π rad) and `unit:REV-PER-SEC`. Whether a
"cycle" is inherently rotational is acknowledged as debatable; the rationale is
documented in the `dcterms:description` of both `unit:CYC` and
`unit:CYC-PER-SEC`.


steveraysteveray:
### edv factor-unit consistency check + edv assignment corrections

**New QA infrastructure**

- Added `qfn:unit.eDimVec.calculate` SHACL function that reuses the existing `qfn:unit.dimVec.calculate` (for the 8 SI dimensions) and `qfn:dimVec.getDimensionExponentFromLocalname` (for R and N), avoiding any duplication of the multiply/pow logic.
- Added `src/build/sparql2shacl/extendedDimensionVector/` pipeline. The `validate.ttl` checks that a unit's assigned `qudt:hasExtendedDimensionVector` is consistent with what the factor-unit calculation produces. The `infer.ttl` is an intentional no-op — edv inconsistencies must be corrected in source, not silently overridden.
- Wired the edv consistency validate step into `pom.xml` after the existing factorUnits validation step.

**edv assignment corrections (units)**

- SR/angle compounds (BARN-PER-SR, M2-PER-SR, PER-SR, W-PER-M2-M-SR, etc.): corrected to R=−2
- M2-SR: corrected to R=+2
- BQ-derived compounds (BQ-PER-KiloGM, BQ-PER-L, CI-PER-KiloGM, etc.): corrected to N=1
- LM corrected to R=2 (lumen = cd·sr); all LM-derived photometric units (LUX, LM-PER-W, LM-HR, LM-SEC, LM-PER-FT2, KiloLM) updated to R=2; CD-PER-LM and CD-PER-KiloLM corrected to R=−2
- RAD-M2-PER-KiloGM, RAD-M2-PER-MOL (optical rotation): corrected to R=1
- /RAD units (M-PER-RAD, PER-RAD, SEC-PER-M3-RAD, SEC-PER-RAD-M3): corrected to R=−1
- M2-PER-HZ-DEG, M2-SEC-PER-RAD, IN-PER-REV, PERCENT-PER-DEG: corrected to R=−1
- NUM-PER-* compounds (NUM-PER-GM, NUM-PER-HA, NUM-PER-M2, etc.): corrected to N=1
- *-PER-NUM compounds (DAY-PER-NUM, HR-PER-NUM, SEC-PER-NUM, etc.): corrected to N=−1
- LUX-HR, LUX-SEC: corrected to R=2
- W-PER-M2-MicroM: corrected from R=−2 to R=0 (no SR in formula)

**unit:NUM, unit:COUNT, unit:BIT**

- `unit:NUM` and `unit:COUNT` are now N=1 (they are counting units).
- 17 dimensionless-ratio and information units (`PERCENT`, `PPM`, `PPB`, `PERMILLE`, `BIT`, `BAN`, `HART`, `SHANNON`, etc.) had their `qudt:scalingOf unit:NUM` redirected to `qudt:scalingOf unit:UNITLESS`. This prevents inheriting N=1 while still satisfying the SHACL rule that units with non-unit conversionMultipliers must have a scalingOf or factorUnit anchor.
- `qudt:CountingUnit` removed from `unit:BIT`, `unit:BYTE`, and `unit:FRACTION`.
- `unit:BIT` retains N=0. It represents the information-entropy bit (conversionMultiplier = ln 2), not the count-of-binary-digits bit. A note has been added to its description; a count-of-binary-digits unit (N=1) is deferred to a future version.

**Remaining warnings (still needs discussion)**

The consistency check produces 76 `sh:Warning` results on every build. All are documented in `DESIGN-NOTES.md` as false positives:

- `unit:RAD` (R=1) and `unit:SR` (R=2): definitional assignments; factor units are dimensionless and compute R=0.
- `unit:BQ` (N=1): radioactive-decay convention; factor units compute N=0.
- `unit:CYC-PER-SEC` (N=0): HZ-equivalence convention; CYC factor computes N=1.
- Full N-M torque family, angular momentum units, moment-of-inertia units, and their angle-divided cascades: Group 2 dimensional-cancellation, R assigned by physical convention.
- PERCENT-PER-HUNDRED/THOUSAND/etc.: HUNDRED and THOUSAND are count units (N=1), so division computes N=−1, but the intended result is N=0 (dimensionless ratio).
- PA-M0dot5, MegaPA-M0dot5, J-PER-M2-SEC0dot5-K, N-M-PER-W0dot5: fractional factor-unit exponents; `qfn:unit.eDimVec.calculate` cannot produce a valid integer-exponent edv IRI for these.

fkleedorfer:
>  no two quantity kinds in QUDT's physical vocabulary are ambiguous solely because of a missing currency dimension.

Let's see:

```
quantitykind:InverseEnergy
  qudt:hasDimensionVector qkdv:A0E0L-2I0M-1H0T2D0 ;

quantitykind:CostPerEnergy
  qudt:hasDimensionVector qkdv:A0E0L-2I0M-1H0T2D0 ;

unit:CCY_EUR-PER-W-HR
  qudt:hasDimensionVector qkdv:A0E0L-2I0M-1H0T2D0 ;
  qudt:hasQuantityKind quantitykind:CostPerEnergy ;
```

Here, the unit could be one of the two quantity kinds, and nothing except common sense can disambiguate. These are the cases for which, if your calculation comes up with a dimension vector for inverse energy, you might just conclude it's in `€/Wh` and be called cost per energy, which would be a surprising result and undermine trust in QUDT. If your unit really does include the currency, you'd be equally surprised to see `/Wh`, or it be called inverse Energy. Downstream systems have to code around these cases with expensive and potentially buggy heuristics. 

If we had a dimension for Currency, there would not be ambiguity in such cases.

This is the only example in our vocabulary because somehow, we have some `qk:CostPerX` without the corresponding `qk:InverseX`, except for X=Energy. However, this is not a pathological example. As we continue to make new quantitykinds and units, there will be more such cases.

Consequently, I am in favor of a currency dimension.

steveraysteveray:
@fkleedorfer, my take is that a QUDT Quantity instance (i.e. a specification or measurement) should always have a stated Unit and a stated QuantityKind. In this example, the unit would be unit:CCY_EUR-PER-W-HR and the QuantityKind would be quantitykind:CostPerEnergy. There is no ambiguity here. This is similar to other cases where the identity of the QuantityKind, not the dimension vector, serves to disambiguate the meaning of the Quantity. This happens frequently. Consider measurements of electrical power, including ActivePower, ReactivePower, ApparentPower, etc. They all share the same dimension vector. Indeed all quantity kinds that share a skos:broader relation by definition have the same dimension vector, and yet they define different contexts. Adding a currency dimension vector does not solve this problem, and indeed is more like an attempt to encroach on the whole reason to have a quantity kind associated with a given Quantity.

steveraysteveray:
Further thinking about this, concerning the addition of a Count dimension...

1. Do N=1 and N=0 units with the same SI DV have different reference units for their conversion multipliers?

No. unit:COUNT (N=1) and unit:UNITLESS (N=0) both have convMult 1.0 against the same reference. unit:NUM-PER-SEC (N=1) and unit:HZ (N=0) both have convMult 1.0 against the same reference (1/s). There is no 2π-like factor separating the N branches. N=1 and N=0 units with the same SI DV are arithmetically identical under conversion.

2. Does N create incommensurable reference unit systems?

No. Unlike R — where unit:RAD-PER-SEC and unit:HZ are genuinely different physical quantities despite identical SI DVs and identical convMults (i.e. 1.0 in this case), and using the wrong reference gives numerically wrong results. In contrast, "count" and a "pure dimensionless quantity" of the same magnitude are arithmetically interchangeable. 5 counts and 5 (dimensionless) are the same number. No conversion error is possible.

3. Is N needed for consistency checking?

Only in the sense that it currently flags units that span count QKs and pure-dimensionless QKs. But that check is equivalent to asking "does this unit inappropriately bridge two semantically distinct QK families?" — which is a property of the QK taxonomy, not of a dimension. After the cleanup work in PR #1461, no unit spans both families anyway, so the check passes vacuously.

**Conclusion**:

N is fundamentally unlike R. R has genuine arithmetic content — it determines which reference unit system a conversionMultiplier is expressed against, and getting it wrong gives numerically wrong conversions. N has no arithmetic content at all. The count/dimensionless distinction is purely semantic, which is precisely the job of the QuantityKind. The argument I'm making against currency applies equally to N: the QK is the right home for this distinction, not a pseudo-physical dimension.

R earns its place in the EDV. N doesn't, by the same standard.

If we agree with this conclusion, then I would remove N from the EDV and just have R. This is actually compatible with one of the two positions in the SI discussions.

steveraysteveray:
@ralphtq and @steveraysteveray are now of the view that a simpler, elegant solution is to handle two of the original extended dimension vector entries with type declarations. This is already in place with qudt:CountingUnit and qudt:CurrencyUnit. Angular concepts are not so simple because there are R2, R-2, R1 and R-1 dimensionalities in the real world, such qk:MomentOfInertia, unit:KiloGM-M2. Just using typing would not support disambiguation of such entities.

This conclusion affects PR 1460 and 1461. Bottom line, the single angular dimension is sufficient in the extended dimension vector, and the other two are easily identified through their rdf:type. There will be a new PR that implements this suggestion.

Edit: The comment below modifies the position of @steveraysteveray

steveraysteveray:
@fkleedorfer, @jhodgesatmb, @ralphtq.

Digging deeper:

For quantity kinds, the counting nature can be expressed by hierarchy position within the skos:broader hierarchy. However, at present there are already three natural groupings of "count-related" quantity kinds:

Pure counts → Count
Count-per-time → CountRate (which itself sits under Frequency)
Count-per-volume → NumberDensity (which sits under InverseVolume)

These three QKs become the effective "counting anchors" in the hierarchy. The N dimension was the only way to express "these three families are all fundamentally counting quantities" in a single machine-readable fact without violating the equal SI-dv constraint on skos:broader. Removing it means that commonality is lost at the QK level — it remains only implicitly, via human understanding of what CountRate and NumberDensity represent.

The practical question is whether that loss matters for use cases. If the main consumer of the N dimension was unit matching/disambiguation (distinguishing e.g. unit:BQ from unit:HZ despite both being T⁻¹), that is handled by a qudt:CountingUnit. If there are query or reasoning patterns that would specifically exploit the N dimension on QKs, those cases are somewhat less elegantly handled by the skos:broader handling.

More generally, this behavior can also apply to currencies, if we continue to include compound entities such as CostPerPower, CostPerEnergy and associated units like unit:CCY_EUR-PER-KiloW. An extended dimension vector that addresses both the "currency" nature and the denominators in this case is a good argument for a single place to check commensurability - namely an extended dimension vector that includes Count, Angularity, and Currency - which was the original proposal.

Please consider and weigh in on whether we should go with an extended dimension vector of:
1. Only R (angle)
2. R and N (angle and count)
3. R, N and C (angle, count, currency)

steveraysteveray:
Superseded by #1469

PR #1464, opened 2026-05-20
Add quantitykind:LiquidLevel

ektrah:
I'm trying to identify the Quantity Kind that is measured/alarmed by these Brick points:

| @id | ofSubstance | hasQuantityKind | subClassOf |
|:-|:-|:-|:-|
| brick:Collection_Basin_Water_Level_Alarm | Collection_Basin_Water | ? | Alarm |
| brick:Collection_Basin_Water_Level_Sensor | Collection_Basin_Water | ? | Sensor |
| brick:Deionised_Water_Level_Sensor | Deionised_Water | ? | Sensor |
| brick:Max_Water_Level_Alarm | Water | ? | Max_Alarm |
| brick:Min_Water_Level_Alarm | Water | ? | Min_Alarm |
| brick:Refrigerant_Level_Sensor | Refrigerant | ? | Sensor |
| brick:Water_Level_Alarm | Water | ? | Alarm |
| brick:Water_Level_Sensor | Water | ? | Sensor |

Both [`quantitykind:Height`](http://qudt.org/vocab/quantitykind/Height) and [`quantitykind:Depth`](http://qudt.org/vocab/quantitykind/Depth) don't seem 100% appropriate (though `quantitykind:Depth` is very close).

So I'm proposing a new Quantity Kind: `quantitykind:LiquidLevel`

/cc @gtfierro

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 - Added  quantitykind:ApparentEnergy and completed the ElectricEnergy specialization hierarchy
 - Added unit:BAR-PER-M (Bar per Metre), unit:PSI-PER-FT (Psi per Foot), and unit:PSI-PER-M (Psi per Metre) for pressure gradients
 - Added 6 energy intensity units: W-HR-PER-FT2, KiloW-HR-PER-FT2, MegaW-HR-PER-FT2, GigaW-HR-PER-FT2, MegaW-HR-PER-M2, GigaW-HR-PER-M2
+- Added quantitykind:LiquidLevel
 
 ### Changed
 
--- a/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
+++ b/src/main/rdf/vocab/quantitykinds/VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl
@@ -7466,6 +7466,16 @@ quantitykind:LinkedFlux
   rdfs:label "Linked Flux"@en ;
   skos:broader quantitykind:MagneticFlux .
 
+quantitykind:LiquidLevel
+  a qudt:QuantityKind ;
+  dcterms:description "Liquid level refers to the measurement of the height of a liquid within a container."^^qudt:LatexString ;
+  qudt:hasDimensionVector qkdv:A0E0L1I0M0H0T0D0 ;
+  qudt:plainTextDescription "Liquid level refers to the measurement of the height of a liquid within a container." ;
+  qudt:wikidataMatch <http://www.wikidata.org/entity/Q1481422> ;
+  rdfs:isDefinedBy <http://qudt.org/$$QUDT_VERSION$$/vocab/quantitykind> ;
+  rdfs:label "Liquid Level"@en ;
+  skos:broader quantitykind:Length .
+
 quantitykind:LiquidVolume
   a qudt:QuantityKind ;
   dcterms:description "Liquid volume is the volume of a given amount of liquid, that is, the amount of space a liquid takes up. There are a number of different units used to measure liquid volume, but most of them fall under either the metric system of measurement or the Imperial system of measurement."^^qudt:LatexString ;


dr-shorthair:
As with all spatial measures, this is not a _quantity_, it is a measure of _position_ which is usually expressed as a _coordinate_ in a _coordinate reference system_ - 1-D in the case of level, altitude, depth or elevation. It is necessary to specify the reference height (vertical datum) which depends on location (typically the geoid or mean-sea-level, ...) and also the measurement direction (up or down). 

The matter of whether QUDT should spill over into spatial comes up from time to time e.g. #379. Doing it right is a big job, which is already the focus of other standards organizations. Please browse the discussion in #379.

steveraysteveray:
@dr-shorthair, just to avoid confusion, your earlier assertion should be written:
As with all spatial measures, this is not a _**quantitykind**_, it is a measure of position...

as we use the term. Correct?

And in fact, perhaps this could offer an avenue for a resolution. Could the specification of the reference datum could be in the qudt:Quantity instance? The vocabulary of qudt:Quantity instances is not managed by QUDT at all. 

In my optimistic view of the world, perhaps we could have some "incompletely resolved" QuantityKind instances like Height, Depth and even LiquidLevel, while acknowledging that the full context must be provided. One can never know in general where or how a height is measured. The context lies within the qudt:Quantity (i.e. within the application using the general concepts of Height, or LiquidLevel). Does this sound viable?

dr-shorthair:
Yes, the proposal relates to a `QuantityKind`. But this would imply individuals with type `Quantity` which is strictly wrong for the reasons canvassed repeatedly. 

Nevertheless, I agree that _scalar_ (1-D) coordinates or positions are implicitly in scope in QUDT because of the conversionOffset+conversionMultiplier structure, which is exemplified in Celsius/Fahrenheit temperature scales. The fact that the conversionOffset is position dependent for Level/Altitude/Depth measures is the give-away of course. 

As for your proposal, I agree that the datum indicator probably belongs on Quantity instances not QuantityKinds. I'd be looking for something like `hasDatum` in the context of such Quantities.

steveraysteveray:
qudt:Quantity currently has three properties:
```
  sh:property qudt:Quantity-hasQuantityKind ;
  sh:property qudt:Quantity-isDeltaQuantity ;
  sh:property qudt:Quantity-quantityValue
```
I can see the value of an additional optional property. I'm now wondering what the value of qudt:hasDatum would be. Maybe this is where QUDT should refer to something from your world, so it should be of type xsd:anyURI? If that works, can you point us to a resource? (This might also be a case where qudt:isDeltaQuantity should be set to true).

steveraysteveray:
I still haven't addressed the original suggestion by @ektrah. I was distracted by thinking about Altitude and geoscience. In the case of a liquid level, we don't have a corpus of geoscience coordinate systems modeling to draw upon. Should we instead just rely on the application developer to define the context and calculations needed, or is there something more we can provide in terms of available properties on qudt:Quantity?

steveraysteveray (approved):
After more thinking, I agree with you that this is a practical solution to your needs, without having to solve the much broader coordinate system / geoscience discussion. So I'm approving, but waiting to hear from other QUDT authors for their opinions before merging.

dr-shorthair:
> I'm now wondering what the value of qudt:hasDatum would be. Maybe this is where QUDT should refer to something from your world, so it should be of type xsd:anyURI? If that works, can you point us to a resource?

The most widely used CRS registry is maintained by [EPSG](https://epsg.org). 
You can browse the contents here: https://epsg.org/search/by-name?sessionkey=77mle77uw2&searchedTerms=vertical

The API is described here: https://apps.epsg.org/api/swagger/index.html
There are both _Datum_ and _VerticalCoordRefSystem_ registers. 
The latter adds direction and units-of-measure to the vertical datum. 

e.g.
https://apps.epsg.org/api/v1/Datum/1292 (Australian vertical working surface 2020)
https://apps.epsg.org/api/v1/Datum/5158 (Bluff 1955)
https://apps.epsg.org/api/v1/CoordRefSystem/5760  (New Zealand - South Island - Bluff vcrs)
https://apps.epsg.org/api/v1/VerticalCoordRefSystem/5760 (ditto)
https://apps.epsg.org/api/v1/VerticalCoordRefSystem/6358 (NAVD88 depth (ftUS))

dr-shorthair:
Or from a better URL (result in XML)
- https://www.opengis.net/def/datum/EPSG/0/1292
- https://www.opengis.net/def/datum/EPSG/0/5158
- https://www.opengis.net/def/crs/EPSG/0/5760
- https://www.opengis.net/def/crs/EPSG/0/6358

fkleedorfer (approved):

ralphtq (approved):

PR #1467, opened 2026-05-26
Srr fix onstate offstate missing triple

steveraysteveray:


fkleedorfer (approved):

src/main/rdf/vocab/types/VOCAB_QUDT-DATATYPES.ttl
```diff
@@ -808,8 +809,9 @@ datatype:ON
 
 datatype:ONstate
   a qudt:OnOffStateTypeEnumeration ;
-  dtype:code "0" ;
+  dtype:code "1" ;
```
ektrah, line 812: @steveraysteveray  In the `OffOnStateTypeEnumeration`, "off" is code "0". So, in the `OnOffStateTypeEnumeration`, "on" should be code "0" as well, no?

PR #1471, opened 2026-05-29
Introduce qudt:eclassCode

ektrah:
* Adds `qudt:eclassCode` in the same style as `qudt:iec61360Code`
* Changes the predicate to `qudt:eclassCode` in all triples where the predicate is `qudt:symbol` and the value is an ECLASS IRDI


steveraysteveray (changes requested):
Looks good, except please don't modify src/main/rdf/schema/SCHEMA_QUDT.ttl. That file will be automatically derived from the SHACL schema file.

steveraysteveray (approved):

PR #1469, opened 2026-05-27
Full extended dimensions

steveraysteveray:
## Summary

This PR builds on PR #1461 and implements Option 3 described [here](https://github.com/qudt/qudt-public-repo/pull/1460#issuecomment-4555265961). It adds a **C (currency) dimension** to the QUDT extended dimension vector (EDV) system, completing the 11-component fixed pattern `A{a}E{e}L{l}I{i}M{m}H{h}T{t}D{d}R{r}N{n}C{c}`.

### Schema changes
- **`SCHEMA_QUDT_NoOWL.ttl`**: Add SHACL property shape `qudt:ExtendedDimensionVector-dimensionExponentForCurrency` (`sh:minCount 1`, `sh:maxCount 1`) to the `qudt:ExtendedDimensionVector` node shape.
- **`infer-dv-exponents.ru`**: Extend the IRI-parsing SPARQL update to extract C; fix the N extraction regex (N is no longer terminal); add C to the extended-dimensions INSERT block.

### Vocabulary changes
- **`VOCAB_QUDT-EXTENDED-DIMENSION-VECTORS.ttl`**: Rename all 286 existing EDV instances to append `C0`; add 6 new `C=1` instances for the currency quantity-kind patterns; add `qudt:dimensionExponentForCurrency 0` explicitly to `edv:NotApplicable`. Total: 293 instances.
- **`VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl`**: Global `C0` update to all 1 213 `qudt:hasExtendedDimensionVector` triples; set `C=1` for `Currency`, `CurrencyPerFlight`, `CostPerEnergy`, `CostPerPower`, `CostPerArea`, `CostPerMass`; add new `quantitykind:CurrencyPerTime`; fix 32 references that incorrectly read `edv:NotApplicableC0`.
- **`VOCAB_QUDT-UNITS-ALL.ttl`**: Global `C0` update to all 2 914 `qudt:hasExtendedDimensionVector` triples; set `C=1` for all 203 `CCY_*` / `MegaCCY_*` units; fix `unit:MegaCCY_USD-PER-YR` quantity kind from `quantitykind:Unknown` to `quantitykind:CurrencyPerTime`.

### Design notes
- **`DESIGN-NOTES.md`**: Document the C dimension (intro, dividing-line section, worked-examples table); fix the namespace prefix on `ExtendedDVConsistencyRule` from `qudt:` to `qfn:`.

### C=1 assignment rules
[...]

PR #1475, opened 2026-06-10
Srr commensurability

steveraysteveray:
This PR builds on top of PR #1469 by merging commensurate quantity kind families that contain members related by qudt:exactMatch. This gives a much more complete picture of commensurability. I will be adding a comment in [our Discussion](https://github.com/qudt/qudt-public-repo/discussions/1473) explaining what I mean by this.

This same set of commits has also been layered on top of PR #1474, so we still have the freedom to choose our strategic choice regarding dimension vectors.


steveraysteveray:
Chose just the SI dimension vector for now.

PR #1476, opened 2026-06-14
Dimensionless decoupling + explicit quantity-kind/unit relations

steveraysteveray:
Two related pieces of work on the quantity-kind/unit model, both **fully backward-compatible** (additive; released graphs unchanged for existing consumers).

## Part 1 — Decouple the dimensionless quantity kinds from commensurability
The abstract grouping nodes (`Dimensionless`, `DimensionlessRatio`, …) were linked by `skos:broader`, making ~90 dimensionless quantity kinds one commensurability family — wrongly reporting e.g. strain, efficiency, Mach number and quantum numbers as mutually comparable. Those 89 edges are now `qudt:organizedUnder` (a grouping relation that does **not** assert commensurability), plus a same-dimension QA guard and a grab-bag unit rehoming. Applicable units are unchanged.

## Part 2 — Explicit quantity-kind/unit relations (was #1479)
Replaces the overloaded use of `skos:broader` and `qudt:hasQuantityKind` with precise, individually-validatable sub-properties; the old property is materialised by the build as a super-property for backward compatibility.

| | commensurate | non-commensurate |
|---|---|---|
| unit → qk | `qudt:unitForQuantityKind` | `qudt:categorizedByQuantityKind` |
| qk → qk | `qudt:specializationOf` | `qudt:organizedUnder` |

- **Declared** the three new sub-properties + SHACL property shapes.
- **Build materialization** regenerates `skos:broader` / `qudt:hasQuantityKind` into dist (and removes `hasQuantityKind` from the src↔dist property-sync, since nothing infers it).
- **Auto-confident re-typings:** 228 `skos:broader`→`specializationOf`, 2598 `hasQuantityKind`→`unitForQuantityKind`, 454 Unknown→`categorizedByQuantityKind`.

**Backward compatibility proven:** built dist is byte-identical to baseline for `skos:broader` (480), `hasQuantityKind` (4088) and `applicableUnit` (15244) pairs — only new sub-property triples added. `mvn clean install` green.

**Deferred (not in this PR):** coincidence-prone hierarchy edges, generic dimensionless units, cross-family unit links, physical constants; and the final behavior-changing step (moving inference onto the new relations + SHACL violations + the `organizedUnder` flip).


src/main/rdf/schema/shacl/SCHEMA_QUDT_NoOWL.ttl
```diff
@@ -1661,6 +1662,21 @@ qudt:order
   rdfs:isDefinedBy <http://qudt.org/$$QUDT_VERSION$$/schema/shacl/qudt> ;
   rdfs:label "order" .
 
+qudt:organizedUnder
+  a rdf:Property ;
+  dcterms:description """
```
ralphtq, line 1682: Move description to the property shape? That way the predicate is reusable.

fkleedorfer (approved):

steveraysteveray:
Superseded

PR #1479, opened 2026-06-16
Explicit quantity-kind/unit relations (specializationOf, unitForQuantityKind, categorizedByQuantityKind)

steveraysteveray:
Introduces explicit, individually-validatable relations to replace the **overloaded** use of
`skos:broader` (quantity kind → quantity kind) and `qudt:hasQuantityKind` (unit → quantity kind),
whose QUDT-intended meaning is precise but which consumers read loosely. This is the first,
**fully backward-compatible** slice of the relations remodel; it is additive and *no worse than
today* for every existing consumer.

Companion app change: qudt-r2 #324 (subtree viewer follows `qudt:specializationOf`).

## The model

Two precise relations on each side; the long-standing property becomes a **materialized
super-property** (kept in releases for backward compatibility) while authors assert the precise
sub-property in source.

| | commensurate (same family → inter-convertible units) | non-commensurate (categorical grouping) |
|---|---|---|
| **unit → quantity kind** | `qudt:unitForQuantityKind` | `qudt:categorizedByQuantityKind` |
| **quantity kind → quantity kind** | `qudt:specializationOf` | `qudt:organizedUnder` *(already present)* |
| **materialized super (not authored)** | — | `qudt:hasQuantityKind` · `skos:broader` |

- **`qudt:specializationOf`** (⊑ `skos:broader`) — the child *is the same quantity kind* as the
  parent, more specific; values comparable/addable and units inter-convertible. Commensurability
  families = connected components under `(specializationOf | exactMatch)`.
- **`qudt:unitForQuantityKind`** (⊑ `hasQuantityKind`) — the unit measures a member of a
  commensurate family it is convertible within.
- **`qudt:categorizedByQuantityKind`** (⊑ `hasQuantityKind`) — the unit is filed under a
  non-commensurate bucket (e.g. `quantitykind:Unknown`, `DimensionlessRatio`) with no
  convertibility claim. Needed because a shared unit (e.g. `PERCENT`) does **not** imply the
  quantity kinds are commensurable.
- `qudt:organizedUnder` is deliberately **left independent of `skos:broader`** for now — making it
  a sub-property would re-materialize `skos:broader` for the dimensionless-decoupled edges and
  re-break commensurability while inference still runs on `skos:broader`. That flip is reserved for
  the final PR.

## Commits (each independently additive)

1. **PR1 — declare the relations** (`f70579e4`): the three sub-properties + their SHACL property
   shapes, registered on the `AbstractQuantityKind` / `Unit` node shapes. Schema only; no data,
   no inference change.
2. **PR2 — build materialization** (`d61419d3`): `src/build/inference/materializeRelationSuperProperties.ttl`
   (`sh:SPARQLRule`, `sh:targetSubjectsOf`) wired as two `<shaclInfer>` steps before
   `applicableUnit`/`broaderAcrossExactMatch`, so released graphs keep `skos:broader` /
   `hasQuantityKind`. **Build-reviewer note:** removed `qudt:hasQuantityKind` from the src↔dist
   property-sync lists (6 occurrences across the duplicated core/extension pipelines) — nothing in
   the build infers `hasQuantityKind` (only authored + read), so the sync was a no-op for it;
   dropping it lets `hasQuantityKind` be dist-only (materialized), symmetric with `skos:broader`.
3. **PR3 — re-type the auto-confident set** (`6a6d0f7e`): 228 single-family `skos:broader` →
   `specializationOf`; 2598 `hasQuantityKind` → `unitForQuantityKind` (single-QK + multi-QK whose
   kinds form one `(skos:broader|exactMatch)` family); 454 `hasQuantityKind quantitykind:Unknown` →
   `categorizedByQuantityKind`.
4. **rename** (`c61b7351`): `measuresQuantityKind` → `unitForQuantityKind`, `categorizedBy` →
   `categorizedByQuantityKind` per review, across schema/shapes/rule/pom/data + labels + descriptions.

## Backward compatibility — proven

Despite re-typing 3280 source triples, the built dist is **byte-identical** to the pre-change
baseline for every legacy predicate (only the new sub-property triples are added):

| predicate | pairs | result |
|---|---:|---|
| `skos:broader` | 480 | identical |
| `hasQuantityKind` | 4088 | identical |
| `applicableUnit` | 15244 | identical |

`mvn clean install` → **BUILD SUCCESS**.

## Scope — what's deferred (needs ontology review, not in this PR)

- 159 coincidence-prone `skos:broader` edges (e.g. the `T⁻¹` rate zoo under `InverseTime`)
- 169 generic-dimensionless unit links and 772 cross-family multi-QK links
- **Physical constants** keep `hasQuantityKind` — they're valued quantities ("value of kind X"),
  nothing infers from them, and ~79/330 point at buckets; not re-typed.

## Not in this PR (the final step)

Moving commensurability / `applicableUnit` **inference** onto the new relations, adding the
same-dimension / numerator-denominator SHACL violations, and flipping `organizedUnder` to a
`skos:broader` sub-property. Those are the only behavior-changing steps and ship last, after the
deferred re-typings, so each interim release stays no-worse.

Detailed audits/checklists (per-edge classification, the dimensionless reclassification, the
release-sequencing plan) are maintained outside the repo and can be shared on request.


src/main/rdf/schema/shacl/SCHEMA_QUDT_NoOWL.ttl
```diff
@@ -39,6 +39,7 @@ qudt:AbstractQuantityKind
   sh:property qudt:AbstractQuantityKind-broader ;
   sh:property qudt:AbstractQuantityKind-latexSymbol ;
   sh:property qudt:AbstractQuantityKind-organizedUnder ;
```
ralphtq, line 41: We should rename the property shapes to use categorizedByQuantityKind instead of 'organizedUnder'. This can be a future PR.

ralphtq (approved):

src/main/rdf/schema/shacl/SCHEMA_QUDT_NoOWL.ttl
steveraysteveray: Happy to discuss renaming organizedUnder, especially if we can settle on something before I publish the next Release. I would rather postpone that release than change the property name after that. 

One big problem with reusing categorizedByQuantityKind the way you suggest is because it can no longer be a subProperty of skos:broader. We would either need it to be a subProperty of skos:broader AND qudt:hasQuantityKind, which is not workable. Or, when re-materializing those two super-properties, we would have to go through a much less elegant conditional process.

So, I'm open to renaming, but not in favor of reusing the other name.

PR #1482, opened 2026-06-18
Fix applicableUnit tie-breaking; add specializationOf/organizedUnder disjointness validation

steveraysteveray:
## Context

PR #1481 ("Move commensurability inference onto the precise relations") introduced `qudt:specializationOf` and `qudt:organizedUnder` as distinct subproperties of `skos:broader`, replacing the single coarse `skos:broader` edge for quantity-kind hierarchy. Both relations still materialise to `skos:broader` via `materializeRelationSuperProperties.ttl`, but they carry different semantics:

- **`qudt:specializationOf`** — the child is commensurable with its parent; unit conversion is licensed within the resulting family.
- **`qudt:organizedUnder`** — the child is categorically related to the parent but is **not** commensurable; no unit conversion is implied.

Two latent defects became apparent once these precise relations were in place.

### 1. `src/build/inference/inferApplicableUnits.ttl` — deterministic union on multi-parent ties

The previous algorithm walked `(skos:broader|qudt:organizedUnder)*` upward, found the nearest unit-bearing ancestor by `ORDER BY ASC(?steps) LIMIT 1`, and emitted that single ancestor's units. `LIMIT 1` is non-deterministic when two parents sit at the same minimum distance and both carry units: the SPARQL engine picks one arbitrarily.

**Live example:** `quantitykind:SignalStrength` has `qudt:specializationOf` edges to both `quantitykind:ElectricField` (3 units) and `quantitykind:ElectricFieldStrength` (10 units) — independent roots at distance 1. The build arbitrarily chose `ElectricField`'s 3 units, silently dropping the other 10.

**Fix:** Replace `ORDER BY ASC(?steps) LIMIT 1` with a two-subquery pattern:

1. Compute `MIN(?steps)` across all unit-bearing non-deprecated ancestors.
2. Keep every ancestor at exactly that minimum distance, then union their units.

For the common case (unique nearest ancestor) the result is identical. For tied ancestors the units are unioned deterministically. Quantity kinds with no unit-bearing ancestors continue to produce no `applicableUnit` triples.

The precise `specializationOf`/`organizedUnder` split that PR #1481 introduced makes this tie scenario more likely going forward (quantity kinds will increasingly have multiple typed parents at the same depth), so fixing the algorithm now avoids silent build-order sensitivity.

### 2. `src/main/rdf/validation/COLLECTION_QUDT_QA_TESTS_ALL.ttl` — disjointness constraint

Because `specializationOf` and `organizedUnder` assert contradictory things about commensurability, a quantity kind must use **at most one** of them as an outgoing relation type. No constraint previously enforced this.

Added `qudt:SpecializationOfAndOrganizedUnderDisjointShape` (`sh:Violation`) that fires when a non-deprecated quantity kind carries both an outgoing `qudt:specializationOf` and an outgoing `qudt:organizedUnder`.

## Test plan

- [ ] `mvn install` passes with no new violations
- [ ] Verify `target/inferred/applicableUnits.ttl` for `quantitykind:SignalStrength` now lists units from **both** `ElectricField` and `ElectricFieldStrength`
- [ ] Confirm no existing quantity kind triggers the new disjointness violation (currently zero have both relation types)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 ### Added
 
+- Added `qudt:SpecializationOfAndOrganizedUnderDisjointShape`, a Violation-severity QA check that fires when a non-deprecated quantity kind carries both an outgoing `qudt:specializationOf` and an outgoing `qudt:organizedUnder`. The two properties are mutually exclusive by design.
 - Added rigor to building the quantity kind commensurability families. Now, a quantity kind family now constitutes the necessary and sufficient requirement for commensurability.
 - Added code to automatically convert labels for quantity kinds into Title Case
 - Added qudt:eclassCode
@@ -32,6 +33,7 @@ and this project is in the process of adopting [Semantic Versioning](https://sem
 
 ### Fixed
 
+- Fixed `inferApplicableUnits` to union units from all equidistant unit-bearing ancestors instead of arbitrarily picking one. The previous `ORDER BY … LIMIT 1` pattern was non-deterministic when two parents sat at the same minimum distance; the new query computes `MIN(?steps)` and keeps every ancestor at that distance, deterministically unioning their units.
 - Fixed a missing qudt:element property for datatype:ONstate and datatype:OFFstate
 
 ## [3.3.0] - 2026-05-25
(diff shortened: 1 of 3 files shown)


ralphtq (approved):
Understand this is in support of PR 1483

PR #1483, opened 2026-06-18
Update QK classifications: hierarchy edges + unitForQuantityKind/categorizedByQuantityKind (§1–§3)

steveraysteveray:
## Background

PR #1481 introduced `qudt:specializationOf`, `qudt:organizedUnder`, `qudt:unitForQuantityKind`, and `qudt:categorizedByQuantityKind` as precise replacements for `skos:broader` and `qudt:hasQuantityKind`. This PR applies those relations throughout the quantity-kind hierarchy and to the first three tiers of unit links.

## Quantity-kind hierarchy changes

New edges added to `VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl`:

- `qudt:specializationOf` — child QKs that are commensurate with (in the same measurement family as) their parent; these share commensurable units
- `qudt:organizedUnder` — child QKs that are grouped under a parent for navigational/categorical purposes but are not commensurate with it (e.g. `ReactivePower organizedUnder ElectricPower`)
- `qudt:exactMatch` — QK pairs that are definitionally equivalent across naming conventions (e.g. `Conductance exactMatch ElectricalConductance`)

Also corrects `rdfs:label` on `quantitykind:ElectricPower`: all non-English labels previously read "active power" rather than "electric power".

## Unit–quantity-kind relation conversions

310 `qudt:hasQuantityKind` links across 223 non-deprecated units have been converted:

| tier | units | links | new predicate | criterion |
|---|---:|---:|---|---|
| §1 Unknown | 14 | 14 | `categorizedByQuantityKind` | target is `quantitykind:Unknown` |
| §2 Dimensionless buckets | 24 | 25 | `categorizedByQuantityKind` | all targets are generic dimensionless placeholders (`Dimensionless`, `DimensionlessRatio`, `Count`, `Ratio`) |
| §3 In-family | 165 | 271 | `unitForQuantityKind` | all QK targets lie in one `(specializationOf ∣ exactMatch)` commensurability family |

Deprecated units are left unchanged with `hasQuantityKind`.

653 `hasQuantityKind` links across 157 units (§4 — cross-family) remain and require individual decisions; those will be addressed in a follow-on commit on this branch.

## What is not in this PR

- §4 unit-level decisions (cross-family links) — in progress
- IEV/ISO/Wikidata references — not added (per project policy, only verified references are included)

PR #1484, opened 2026-06-19
Add equivalent concentration units for clinical chemistry (Eq/L family)

jimkont:
## Summary

Adds eight unit definitions classified under `quantitykind:EquivalentConcentration` for use in clinical chemistry and clinical-trial data exchange:

| Unit | Symbol | Multiplier (to mol/m³) |
|---|---|---|
| `unit:EQ-PER-L` | Eq/L | 1.0E3 |
| `unit:MilliEQ-PER-L` | mEq/L | 1.0E0 |
| `unit:MicroEQ-PER-L` | µEq/L | 1.0E-3 |
| `unit:NanoEQ-PER-L` | nEq/L | 1.0E-6 |
| `unit:MilliEQ-PER-DeciL` | mEq/dL | 1.0E1 |
| `unit:MilliEQ-PER-MilliL` | mEq/mL | 1.0E3 |
| `unit:MicroEQ-PER-MilliL` | µEq/mL | 1.0E0 |
| `unit:NanoEQ-PER-MilliL` | nEq/mL | 1.0E-3 |

## Motivation

These units are pervasive in clinical-trial laboratory data:

- **mEq/L** is the standard reporting unit for serum, plasma, and urine electrolytes (Na⁺, K⁺, Cl⁻, HCO₃⁻, Ca²⁺, Mg²⁺) in U.S. clinical practice and in SDTM-tabulated trial submissions.
- **mEq/mL** appears on intravenous electrolyte concentrate product labels (e.g. potassium chloride concentrate, typically 2 mEq/mL).
- The `µEq` and `nEq` prefixes cover trace measurements; the `Eq/L` base and `mEq/dL` variant complete the prefix family for symmetry.

The vocabulary already contains `unit:MilliEQ` (under `quantitykind:ReactiveCharge`) and `unit:MilliEQ-PER-HectoGM`, but no per-volume equivalent concentration units. This PR fills that gap.

## Prerequisite units

This PR depends on `unit:EQ`, `unit:MilliEQ`, `unit:MicroEQ`, and `unit:NanoEQ` existing as base/prefixed amount units under `quantitykind:ReactiveCharge`. `unit:MilliEQ` is confirmed present; the others need verification. If any are missing, they should be added alongside (each is a single-unit definition following the `unit:MilliEQ` pattern with the appropriate `qudt:prefix` and `qudt:conversionMultiplier`). I can submit the missing ones in a follow-up commit on this PR if maintainers confirm what's needed.

## `applicableSystem`: deliberately omitted

Equivalent is not formally part of SI, CGS, Imperial, USCS, ATOM, PLANCK, or ASU. Some references describe Eq/L as "the SI unit of normality," but the SI Brochure does not list equivalent as a unit, and ISO 80000-9 does not include equivalent concentration. I've omitted `qudt:applicableSystem` rather than make an inaccurate claim; maintainers can add an appropriate system if QUDT chooses to recognise one.

## Conversion semantics (related issue)

These units share the dimension vector `A1E0L-3I0M0H0T0D0` with `quantitykind:Concentration` (mol/L, mmol/L, etc.). The User Guide's [section 5 conversion pattern](https://github.com/qudt/qudt-public-repo/wiki/User-Guide-for-QUDT#5-unit-conversion-example-in-qudt) will therefore happily return a "convert mEq/L to mmol/L by multiplier 1.0" result, which is correct only for monovalent ions (Na⁺, K⁺, Cl⁻). For divalent ions (Ca²⁺, Mg²⁺) the correct factor is 2; the conversion is fundamentally analyte-dependent and cannot be encoded in unit-level multipliers alone.

This is the same class of issue as Torque vs Energy (both N·m, dimensionally identical, semantically distinct), and I've opened a separate issue (https://github.com/qudt/qudt-public-repo/issues/1485) to discuss whether the User Guide should add a caveat or whether QUDT should model "linearly convertible by multiplier" explicitly between quantity kinds. The choice does not block this PR — the units as submitted are individually correct; the question is what guidance to give consumers about cross-quantity-kind conversion.


steveraysteveray (changes requested):
This is a quite timely and thoughtful submission - I thank you for it. First, there is no problem with your new units at present, but as it happens, we are about to merge two updates, #1482 and #1483 that will be addressing precisely your second point about commensurability of units and quantity kinds that share a dimensionality (composition) but not the same context (semantics). I invite you to take a look at the table in #1479 where we did some of the prework, including defining new relations that distinguish the very issues you raise.

Because of this, I would ask you to hold off until we merge these last two PRs, at which point you will need to replace your qudt:hasQuantityKind relations with qudt:unitForQuantityKind. Once you do that, and merge the new main branch (after we merge our two PRs), you should get a clean build once more.

Armed with this new precision, we could then consider the other point you raise, by defining some specialization quantitykinds that would be used in the different valency conditions. I haven't thought deeply about this yet, but I believe their use would be from a qudt:Quantity that would refer to a particular unit, and the appropriate quantitykind for that specific context. Happy to discuss this further!

I will let you know when our new main branch is ready for your adjusted PR.

jimkont:
Thank you @steveraysteveray — great to hear this is already being addressed, and the #1479 relations are exactly the distinction that was missing.
I've updated the PR:

* Switched the eight units from qudt:hasQuantityKind to qudt:unitForQuantityKind quantitykind:EquivalentConcentration — correct here since the Eq units are mutually commensurable within the family (mEq/L ↔ Eq/L ↔ µEq/L are genuine multiplier conversions).
 * Changed quantitykind:EquivalentConcentration from qudt:specializationOf quantitykind:Concentration to qudt:organizedUnder quantitykind:Concentration.

The reasoning for the second change: under the new model specializationOf asserts commensurability (values comparable/addable, units inter-convertible), but EquivalentConcentration → Concentration isn't unconditionally convertible — 1 mEq/L = 1 mmol/L for monovalent ions but 0.5 mmol/L for divalent, so the conversion depends on the analyte's valence rather than a unit-level constant. organizedUnder keeps the conceptual hierarchy and unit-applicability grouping while making no commensurability claim, so the commensurability test correctly treats EquivalentConcentration as its own family and doesn't auto-convert mEq/L ↔ mmol/L. This mirrors the dimensionless decoupling in #1476, where same-dimension but non-commensurable kinds were moved off the commensurable relation for the same reason.

steveraysteveray (approved):
This is excellent! You clearly have a full understanding of what we are trying to capture in our recent modeling upgrades. Hopefully we can continue to refine those thoughts, possibly with valency-specific quantity kinds organizedUnder EquivalentConcentration. We are also contemplating adding a new textual property named something like qudt:conversionDependency that could describe what the factors that might be to support a precise conversion. This could be used not only for your equivance chemistry, but also for other units/quantity kinds that do not lend themselves to linear conversion multipliers.

PR #1492, opened 2026-06-25
Tag untagged rdfs:label values on QuantityKind and Prefix as English

steveraysteveray:
## Summary

- The title-case fix (`mvn -Pfix install`) now defaults an untagged `rdfs:label` to `@en` instead of preserving the missing tag, matching `qudt:RdfsLabelInTitleCaseShapeWarning`, which already treats untagged labels as English for `QuantityKind` and `Prefix`.
- A new dedup rule removes untagged `rdfs:label` values that were redundant leftovers next to an identical, already `en`/`en-US`/`en-UK`-tagged label.
- Deprecated entities are skipped by both rules, as before, so this is enforced automatically going forward rather than needing repeated one-off cleanup.

Net effect on `VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl` and `VOCAB_QUDT-PREFIXES.ttl`: 94 `QuantityKind` labels fixed (62 tagged, 32 deduped) and 33 redundant `Prefix` label duplicates removed. The 4 remaining untagged `QuantityKind` labels all belong to deprecated entities and are intentionally left alone.

## Test plan

- [x] `mvn -Pfix install` — `BUILD SUCCESS`, applied the fix to src
- [x] `mvn install` (clean, no `-Pfix`) — `BUILD SUCCESS`, confirms src/dist are back in sync
- [x] Verified via SPARQL against the built dist graphs that 0 non-deprecated `QuantityKind`/`Prefix` labels remain untagged


qudt:
Auto-approving because this is non-controversial.

PR #1496, opened 2026-07-01
Add BIT_COUNT/BYTE_COUNT counting units for BitDataVolume/ByteDataVolume

steveraysteveray:
## Summary

`unit:BIT` and `unit:BYTE` are information-entropy units (`conversionMultiplier` = ln(2) / 8·ln(2)), not the plain count-of-binary-digits usage used in phrases like "a 64-bit integer" or to express file/data-volume sizes. `unit:BIT`'s own description already flagged this gap. `quantitykind:BitDataVolume` and `quantitykind:ByteDataVolume` existed but had zero `applicableUnit`.

- Added `unit:BIT_COUNT` (`qudt:CountingUnit`, `conversionMultiplier` 1.0) → `unitForQuantityKind BitDataVolume`
- Added `unit:BYTE_COUNT` (`qudt:CountingUnit`, `hasFactorUnit BIT_COUNT` × 8, `conversionMultiplier` 8.0) → `unitForQuantityKind ByteDataVolume`
- `BitDataVolume` and `ByteDataVolume` are now `qudt:organizedUnder quantitykind:Count`
- Cross-linked `unit:BIT`↔`unit:BIT_COUNT` and `unit:BYTE`↔`unit:BYTE_COUNT` with `rdfs:seeAlso`, and updated the `unit:BIT`/`unit:BYTE` descriptions to point to the new counting-sense units

Build (including `-Pfix`) has been run locally by the maintainer and passes.

## Test plan

- [ ] `quantitykind:BitDataVolume` has `applicableUnit unit:BIT_COUNT`
- [ ] `quantitykind:ByteDataVolume` has `applicableUnit unit:BYTE_COUNT`
- [ ] `unit:BYTE_COUNT` conversion multiplier resolves to 8.0 relative to `unit:BIT_COUNT`
- [ ] `unit:BIT`/`unit:BYTE` (entropy units) are unaffected in their existing `InformationEntropy`/`ShannonDiversityIndex` links

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1497, opened 2026-07-02
Redefine BIT/BYTE ladder as counting units; separate from information entropy

steveraysteveray:
## Summary

Supersedes #1496. Rather than adding parallel `_COUNT` units, this redefines `unit:BIT`, `unit:BYTE`, `unit:OCTET`, and the entire SI/binary-prefixed and compound (`-PER-SEC`, `-PER-M`/`M2`/`M3`) ladder in place, since most users think of "bit"/"byte" in the counting/storage sense, not the information-entropy sense.

**Root cause:** `unit:BIT` (conversionMultiplier = ln2 ≈ 0.693) was a near-exact duplicate of the pre-existing `unit:SHANNON` — both scaled units of `unit:NAT`, both linked to `InformationEntropy`/`ShannonDiversityIndex`. Meanwhile `BitDataVolume`/`ByteDataVolume` (the counting-sense quantity kinds) had zero applicable units, and the prefixed bit ladder pointed at yet a third quantity kind (`DatasetOfBits`).

**Changes — `unit:BIT`/`BYTE`/`OCTET` + ~58 prefixed/compound units:**
- `conversionMultiplier` changed from ln2-scaled decimals to clean integer counts: `BIT`=1, `BYTE`=`OCTET`=8, `KiloBYTE`=8000, `GibiBYTE`=8589934592, `ExbiBYTE`=9223372036854775808, etc. Compounds (`-PER-SEC`, `-PER-M{,2,3}`) rescaled to match.
- All volume units now `qudt:CountingUnit`, pointing at `BitDataVolume`/`ByteDataVolume`; density/rate compounds keep their existing `LinearBitDensity`/`AreaBitDensity`/`VolumetricBitDensity`/`DataRate`/`ByteRate`/`BitRate` links.
- `unit:BIT` no longer `scalingOf unit:NAT` and drops the `InformationEntropy`/`ShannonDiversityIndex` links; its description now explains the counting-vs-entropy distinction (ISO/IEC 80000-13) and points to `unit:SHANNON`/`NAT`/`HART`. `unit:BYTE` has a short pointer back to `unit:BIT`. The entropy family itself (`SHANNON`, `NAT`, `HART`, `BAN`) is untouched.

**Quantity kinds:**
- Added `quantitykind:AmountOfData` (`organizedUnder quantitykind:Count`) as the umbrella "amount of digital data" kind.
- `BitDataVolume`, `ByteDataVolume`, `DatasetOfBits` are now `qudt:specializationOf AmountOfData`, keeping bits and bytes mutually commensurable (8:1) even though they point to distinct concrete quantity kinds.
- `BitDataVolume` and `DatasetOfBits` (dimensionally and definitionally identical) are `qudt:exactMatch`.

**Verification done:** both source files re-parse cleanly with rdflib; spot-checked multiplier arithmetic across the full ladder (`KibiBIT`=1024, `GibiBYTE`=8589934592, `ExbiBYTE`=9223372036854775808, …); confirmed no quantity kind carries both `specializationOf` and `organizedUnder` (disjointness rule); confirmed the entropy units (`SHANNON`/`NAT`) are unaffected. Build (`mvn install`, including `-Pfix`) has been run locally by the maintainer and passes clean with no `-Pfix` corrections needed.

## Breaking change / migration

This changes the **value** of previously-released `unit:BIT`, `unit:BYTE`, and the whole prefixed ladder. Anyone using these units for information-entropy calculations should switch to `unit:SHANNON` (base 2), `unit:NAT` (base e), or `unit:HART` (base 10) — see CHANGELOG for the full note.

## Test plan

- [ ] `unit:BIT` conversionMultiplier = 1.0, `unitForQuantityKind` = `BitDataVolume` + `DatasetOfBits`
- [ ] `unit:BYTE`/`unit:OCTET` conversionMultiplier = 8.0, `unitForQuantityKind` = `ByteDataVolume`
- [ ] `unit:SHANNON`/`unit:NAT` unaffected (still `InformationEntropy`/`ShannonDiversityIndex`, ln2/1.0 scaling)
- [ ] `BitDataVolume`, `ByteDataVolume`, `DatasetOfBits` all resolve to the `AmountOfData` commensurability family
- [ ] No `specializationOf`/`organizedUnder` disjointness violations


jhodgesatmb:
I am against changing the definition of unit:BIT. Whether people use BIT as a counting unit, or not, should not be important to us. We should only be concerned with the official definition is/should be. I believe that the technical foundation is that BIT is an information unit.

steveraysteveray:
Officially — in the sense of the governing standard — the bit is a counting/storage unit, and the entropy reading is the deprecated legacy usage. The authority here is ISO/IEC 80000-13 (Quantities and units — Information science and technology), which draws the line explicitly:

Storage capacity / data volume is quantified in bits (with the octet/byte as 8 bits, and the SI and binary prefixes attached to those). This is the counting sense: one bit is one binary storage element.
Information content and entropy get their own units: the shannon (Sh) for base 2, the nat for base e, and the hartley (Hart) for base 10. The standard acknowledges that "bit" has historically been used as a synonym for the shannon — Shannon himself used it that way in 1948 — but deprecates that usage precisely to keep the two senses apart.
The reason the standard splits them is the one that came up when we did the redefinition: the two are not related by a fixed conversion. A stored bit carries at most one shannon of entropy (exactly one only for a uniformly random bit), so treating "bit" as an entropy unit with a hard ln(2)-nat equivalence blurs a data-dependent inequality into a constant. Other unit systems are consistent with the counting reading too — UCUM defines bit as a plain dimensionless unit with value 1, not an entropy measure.

So QUDT's historical position (BIT = ln(2) nat, tied to InformationEntropy) reflected the older Shannon-era convention, and the redefinition we shipped in PR [#1497](https://github.com/qudt/qudt-public-repo/issues/1497) — BIT as a CountingUnit with multiplier 1, entropy served by SHANNON/NAT/HART — is the alignment with the current official standard, not a departure from it. The one caveat worth stating: "official" here means ISO/IEC 80000-13; colloquial usage in information theory literature still says "bits of entropy" routinely, and that usage isn't wrong so much as pre-standard — it's what unit:SHANNON now denotes in QUDT.

jhodgesatmb:
@steveraysteveray quote: “ Storage capacity / data volume is quantified in bits (with the octet/byte as 8 bits, and the SI and binary prefixes attached to those). This is the counting sense: one bit is one binary storage element.”

Does SI say that a bit is a counting unit? Your quote says that one bit is one binary storage element, so one gram is one mass storage element? A gram is not a counting unit. I guess I need to see that SI considers a bit a counting unit, not that anyone else does.

I am not against PRs in general but our definition of bit has been in place a long time and I do not understand why we want to change it.

my comment is only about bit and not about information entropy which I see as different.

steveraysteveray:
Below is a synopsis of the considerations I made while proposing this PR. Note the alternative that was considered (PR 1496) and the review of the standards.

# Redefinition of `unit:BIT` / `unit:BYTE`: rationale and record (PR #1497)

## The problem: one ladder of names served two different quantities

Before #1497, the bit/byte units tried to encode both **data volume** (counting binary digits — "a 64-bit integer", file sizes) and **information entropy** (Shannon's measure). These are related only by a data-dependent inequality — a stored bit carries *at most* one shannon of entropy — so no fixed conversion factor can connect them. The unresolved dual identity produced a cascade of inconsistencies:

1. **`unit:BIT` was an exact duplicate of `unit:SHANNON`** — same `conversionMultiplier` (ln 2 ≈ 0.693147), same `scalingOf unit:NAT`, same quantity kinds (`InformationEntropy`, `ShannonDiversityIndex`) — with no link between them.
2. **Quantity-kind targets were split incoherently across the ladder:**
   - base `BIT`, `BYTE` → `InformationEntropy`
   - all prefixed BYTEs (`KiloBYTE`…`ExbiBYTE`) → `InformationEntropy`
   - most prefixed BITs (`KiloBIT`, `MegaBIT`, `GigaBIT`, `TeraBIT`, `PetaBIT`, `KibiBIT`, `MebiBIT`, `GibiBIT`) → `DatasetOfBits`
   - the remaining prefixed BITs (`ExaBIT`, `TebiBIT`, `PebiBIT`, `ExbiBIT`) → `InformationEntropy` (same ladder, arbitrary split)
   - meanwhile the counting-sense quantity kinds `BitDataVolume` / `ByteDataVolume` had **zero units**, and `DatasetOfBits` duplicated `BitDataVolume` with no link.
3. **`BYTE exactMatch OCTET` pointed at different quantity kinds** (`InformationEntropy` vs `DatasetOfBits`), violating the exactMatch-congruence invariant.
4. **Type assertions contradicted the numbers:** the prefixed BYTEs were typed `qudt:CountingUnit` while carrying transcendental entropy multipliers (`KiloBYTE` = 8000·ln 2 ≈ 5545.18); the equally counting-flavored base units and prefixed BITs were *not* typed `CountingUnit`.
5. **Descriptions said "storage", multipliers said "entropy":** `KiloBYTE` — "a multiple of the unit byte for digital information storage… 1000 bytes" — with a multiplier that asserted every stored bit carries exactly ln 2 nat of entropy. Intra-ladder ratios were right; conversions to `NAT`/`HART` silently hard-coded the maximum-entropy assumption.
6. **Interop mismatch:** `unit:BIT` carried `ucumCode "bit"`, but UCUM defines `bit` = 1 (counting), not 0.693 nat.
7. **The gap was documented but unresolved:** `unit:BIT`'s own description conceded it was the entropy bit, "not the count-of-binary-digits bit used in phrases such as 'a 64-bit integer'", promising a separate counting unit "in a future version".

## Alternative A (examined and abandoned): a parallel `_COUNT` ladder

Implemented as PR #1496 (closed): keep the entropy-scaled ladder untouched, mint `unit:BIT_COUNT` (`CountingUnit`, multiplier 1.0 → `BitDataVolume`) and `unit:BYTE_COUNT` (multiplier 8.0, factor 8 × `BIT_COUNT` → `ByteDataVolume`), modeled on the existing `unit:PIXEL_COUNT`, with `rdfs:seeAlso` cross-links.

**Pros:** fully backward compatible (no released value changes, no migration); no deprecation; honored the fix path documented in `unit:BIT`'s description; small, low-risk diff; followed an accepted naming precedent.

**Cons:**
- The intuitive URI (`unit:BIT`) keeps the minority meaning — mainstream users reaching for it get entropy semantics they don't intend; the wrong default by design.
- Completeness forces ~24–26 prefixed `_COUNT` twins, each sharing a symbol with its entropy sibling (two `kB` units with different multipliers — worse confusion than the gap it fixes).
- Fixes none of the pre-existing incoherence (items 1–6 above all survive).
- Standards/interop stay misaligned (the UCUM code `bit` rightfully belongs to the counting sense and cannot map to two QUDT units).
- Permanent maintenance tax: every future bit/byte unit needs twin minting.

## Alternative B (chosen, PR #1497): redefine the ladder in place as counting units

The decisive observation: the entropy sense **already had a complete, standards-blessed home** — `unit:SHANNON` (base 2), `unit:NAT` (base e), `unit:HART` (base 10) — so preserving `unit:BIT` for entropy preserved nothing but a duplicate of the shannon.

- `BIT` = 1, `BYTE` = `OCTET` = 8, full SI + binary prefix ladder rescaled to exact integer counts; all volume units typed `qudt:CountingUnit` and pointed at `BitDataVolume` / `ByteDataVolume`; compound (`-PER-SEC`, `-PER-M/M2/M3`) units rescaled to match.
- New umbrella `quantitykind:AmountOfData` (`organizedUnder Count`) with `BitDataVolume`, `ByteDataVolume`, `DatasetOfBits` as `specializationOf` (bits and bytes stay mutually commensurable); `BitDataVolume exactMatch DatasetOfBits`. All 27 volume units also point at `AmountOfData` directly, so it carries the full bit+byte set while each specialization keeps only its own base's units (the `ElectricPower` / W-VA-VAR pattern).
- Entropy family (`SHANNON`/`NAT`/`HART`/`BAN`) untouched; `unit:BIT` no longer duplicates the shannon; descriptions on `BIT`/`BYTE` explain the counting-vs-entropy distinction and direct entropy users to the shannon.
- **Breaking change**, handled as a corrective fix with a loud CHANGELOG migration note: consumers using `unit:BIT`/`unit:BYTE` for information entropy switch to `unit:SHANNON` (value-identical to the old `unit:BIT`).

Known follow-ups deliberately not addressed in #1497: `OCTET-PER-SEC` → `BitRate` (a bytes-per-second unit on the bit-rate kind) and the `BIT-PER-SEC`→`DataRate` vs `BYTE-PER-SEC`→`ByteRate` asymmetry.

## What the standards say

**UCUM** (spec §4.6, Table 22) defines the bit twice, explicitly:

> "The bit is defined twice. One definition with a subscript letter 's' is defined as the logarithmus dualis of the number of distinct signals. However this unit can not practically be used to express more than 1000 bits. … Therefore we define a second symbol for bit, without the suffix, to be the dimensionless unit 1."

| Code | Kind of quantity | Definition |
|---|---|---|
| `bit_s` | amount of information | `ld(1 1)` — special logarithmic unit, non-metric |
| `bit` | amount of information | **= 1** (plain dimensionless unit, metric/prefixable) |
| `By` (byte) | amount of information | **= 8 bit** |
| `Bd` (baud) | signal transmission rate | = 1/s |

So UCUM's practical, prefixable `bit` — the one behind `kbit`, `KiBy`, and the `ucumCode "bit"` mapping — is the counting unit = 1.

**IEC 80000-13** (*Information science and technology*; current edition 2025, replacing 2008, which superseded IEC 60027-2:2005 subclauses 3.8/3.9 and IEC 60027-3): storage capacity is a dimension-one quantity with the **bit** as its unit (byte = octet = 8 bit; SI prefixes strictly powers of ten, binary prefixes kibi… for powers of 1024), while **information content/entropy** takes the **shannon (Sh)**, **nat**, and **hartley (Hart)**, with a note deprecating the old use of "bit" as a name for the shannon. *(Paraphrase — the standard is paywalled; the deprecation-note wording is from the 2008 edition and should be verified against the 2025 text before quoting.)*

Both authorities agree: the everyday, prefixable bit is the counting unit; entropy gets its own unit. QUDT pre-#1497 had this inverted; post-#1497 it matches both.

## Sources

- UCUM specification: https://ucum.org/ucum (§4.6, Table 22)
- UCUM spec source (prose quoted above): https://raw.githubusercontent.com/ucum-org/ucum/main/ucum.html
- UCUM machine-readable definitions: https://raw.githubusercontent.com/ucum-org/ucum/main/ucum-essence.xml
- IEC 80000-13:2025: available via the IEC webstore (https://webstore.iec.ch, search "IEC 80000-13")
- Edition/status summary: https://en.wikipedia.org/wiki/ISO/IEC_80000

steveraysteveray:
**Addendum — BIPM's position:** The SI Brochure (9th edition) does not define the bit and does not list it among the non-SI units accepted for use with the SI. The bit appears exactly once in the entire brochure, as the example in the prefix rule (§3, "Decimal multiples and sub-multiples"):

> "The SI prefixes refer strictly to powers of 10. They should not be used to indicate powers of 2 (for example, one kilobit represents 1000 bits and not 1024 bits). The names and symbols for prefixes corresponding to powers of 2 are the following: kibi (Ki) for 2¹⁰, … and quebi (Qi) for 2¹⁰⁰ (see IEC 80000-13 for details)."

BIPM thus takes no formal position on the bit itself, but its only usage is a counting statement (1 kbit = 1000 bit), and it explicitly defers the domain to IEC 80000-13 — the standard whose bit/shannon separation this change implements.

Source: SI Brochure, 9th edition — https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf (§3)

### **What is a "counting unit"?** 
The consensus definition comes from the SI Brochure's treatment of *quantities that are a number of entities*. The brochure distinguishes these from all other quantities of dimension one (§2.3.3):

> "It is especially important to have a clear description of any quantity with the unit one (see section 5.4.7) that is expressed as a ratio of quantities of the same kind (for example length ratios or amount fractions) or as a number of entities (for example number of photons or decays)."

and §5.4.7 explains how named counting units arise:

> "Quantities that are ratios of quantities of the same kind (for example length ratios and amount fractions) have the option of being expressed with units (m/m, mol/mol) … Quantities that are a number of entities do not have this option. Whilst not part of the SI, descriptive terms are commonly used **in place of the unit one to indicate the entities being considered** when expressing the values of quantities that are a number of entities. If this is done, the definitions of such terms and the conventions for their use must be clearly documented."

So a **counting unit** is: a documented, descriptive stand-in for the unit one, whose sole job is to identify *which discrete entity* is being counted; the value of the quantity is the integer (or integer-derived) count of those entities, the dimension is one, and the conversion factor to the elementary entity is exactly 1 (or a fixed integer, e.g. byte = 8 bit). This also explains why such units matter practically: SI prefix symbols "can neither be attached to the symbol 1 nor to the name 'one'" (§5.4.7), so a named counting unit like the bit is what makes "kilobit = 1000 bit" expressible at all.

QUDT's schema definition is consistent with this: `qudt:CountingUnit` — "A Counting Unit is used for all units that express counts. Examples are Atomic Number, Number, Number per Year, Percent and Sample per Second." (subclass of `qudt:DimensionlessUnit`).

Under this definition, the post-#1497 `unit:BIT` (multiplier 1, entity = binary digit) and `unit:BYTE` (= 8 bit) are textbook counting units — while the pre-#1497 entropy-scaled BIT (0.693… nat) was not a counting unit at all, despite the prefixed BYTE ladder having been typed `qudt:CountingUnit`.

Sources: SI Brochure, 9th ed., §2.3.3 and §5.4.7 — https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf ; QUDT schema — http://qudt.org/schema/qudt/CountingUnit

jhodgesatmb:
It seems that associating unit:Bit with InformationEntropy was where things went astray. Information and information entropy seem to be pretty different. In this respect PR #1497 seems like a valid change. I am unfamiliar with the term BIT Ladder so that has thrown me off, but also referring to unit:Bit as a counting unit when it has historically been used to represent information has also been confusing. Are all dimensionless units counting units?

jhodgesatmb (approved):

Issue #1503, opened 2026-07-15
quantitykind:PowerConstant — English description contradicts its dimension; likely the ECLASS twin of ForceConstant


steveraysteveray:
Noticed while closing out the O/P/Q quantity kinds lacking applicable units (#1502). `quantitykind:PowerConstant` was left unit-less there rather than guessed at, because its description and its dimension disagree. Evidence gathered since suggests it is resolvable.

## Current state

```turtle
quantitykind:PowerConstant
  a qudt:QuantityKind ;
  dcterms:description "ratio indicating the relationship between continuous power and continuous current"@en ;
  qudt:eclassCode "0173-1#Z4-BAJ330#003" ;
  qudt:hasDimensionVector qkdv:A0E-1L1I0M1H0T-2D0 ;
  qudt:plainTextDescription "Verhältnis, das den Zusammenhang zwischen der Dauerkraft zum Dauerstrom kennzeichnet"@de ;
  rdfs:label "Power Constant"@en-US .
```

It has no applicable unit.

## The contradiction

- The **English** description says *power ÷ current*. That is watts per ampere = **volts**, dimension `A0E-1L2I0M1H0T-3D0`.
- The asserted **dimension** `A0E-1L1I0M1H0T-2D0` is **N/A** (= T·m = Wb/m = V·s/m) — not volts.

## The German text resolves it

`Dauerkraft` is continuous **force** (*Kraft*), not power (*Leistung*). So the German reads *continuous force ÷ continuous current* = **N/A**, which matches the asserted dimension exactly. Two of the three sources agree; the **English description is the outlier**.

ECLASS is German in origin, so its German text is the authoritative side, and *Kraft* → "power" is a well-known false friend. This suggests the English description — and possibly the `rdfs:label "Power Constant"` itself — is a mistranslation of a force constant.

## Consequence: probable duplicate of ForceConstant

`quantitykind:ForceConstant` is the same quantity, from the other source vocabulary:

| | `PowerConstant` | `ForceConstant` |
|---|---|---|
| source | ECLASS `0173-1#Z4-BAJ330#003` | IEC `0112/2///62720#UAD055` |
| dimension | `A0E-1L1I0M1H0T-2D0` | `A0E-1L1I0M1H0T-2D0` |
| meaning | continuous force ÷ continuous current (per the German) | force ÷ current, F = k·I, N/A |
[...]

PR #1512, opened 2026-07-22
Drop the logarithm base from quantity-kind names where the source vocabularies don't split by base

steveraysteveray:
Resolves the information-content part of #1410; raised by @ektrah in #1503.

## Principle

The logarithm base belongs to the **unit**, not the quantity kind — *except* where a standards body defines the base-specific quantities separately. Every call below is decided by an IEC CDD entry rather than by judgment.

## Merged — IEC treats these as one quantity

[`UAD066` - information content](https://cdd.iec.ch/cdd/iec62720/iec62720.nsf/ListsOfUnitsAllVersions/0112-2---62720%23UAD066?opendocument) lists **shannon, hartley and natural unit of information as three units of a single quantity**. Only ECLASS splits it, and QUDT need not mirror that.

**Removed** (added after 3.4.0, never released — no deprecation needed):
- `quantitykind:BinaryLogarithmicInformationContent`
- `quantitykind:CommonLogarithmicInformationContent`
- `quantitykind:NaturalLogarithmicInformationContent`
- `quantitykind:TimeRelatedNaturalLogarithmicRatio`

**Deprecated in 3.5.0** (these shipped in 3.4.0) in favour of `quantitykind:InformationFlowRate`:
- `quantitykind:BinaryLogarithmicMedianInformationFlow`
- `quantitykind:CommonLogarithmicMedianInformationFlow`
- `quantitykind:NaturalLogarithmicMedianInformationFlow`

"Median" is dropped as a statistical property of a particular measurement rather than of the quantity kind.

`InformationContent` gains ECLASS BAJ463/BAJ468/BAJ469; `InformationFlowRate` gains BAJ464/BAJ470/BAJ471 — each with an `rdfs:comment` recording the many-to-one mapping. Both generics already carried the complete unit set, so **no unit was moved and no new quantity kind was minted**; the eight now-redundant `unitForQuantityKind` links are removed. Nine `dcterms:isReplacedBy` links are repointed so every redirect resolves to a live quantity kind.

> Note on annotation style: `rdfs:comment` is used rather than a Turtle `#` comment, because spotless re-serialises the vocabularies and no `#` comments survive in these files.

## Deliberately NOT merged — IEC splits these on purpose

| IEC entry | Name | Units |
|---|---|---|
| `UAD085` | logarithmic ratio to base 10 | bel, decibel |
| `UAD086` | logarithmic ratio to base e | neper |
| `UAD084` | logarithmic frequency interval to base 10 | decade (ISO 80000-8, 8‑3b) |

Separate entries, separate IEV sources (702‑07‑01 / 702‑07‑03), disjoint unit sets — and `CommonLogarithmicRatio` additionally carries `siExactMatch si-quantity:RLGB`. So `Common`/`NaturalLogarithmicRatio`, `Binary`/`CommonLogarithmicFrequencyInterval` and `CommonLogarithmicRatioPerLength` keep both their splits and their names.

## Also included

Two small corrections that surfaced from the same CDD lookup, kept as separate commits:

- **`CommonLogarithmicRatio` applicable units.** It carried `unit:DECADE`, but the decade is UAD084's unit (the *frequency interval*); UAD085's units are the bel and the decibel. `DECADE` removed, `unit:B`/`unit:DeciB` added — they previously pointed only at the four `Sound*Level` quantity kinds.
- **`quantitykind:Capacity` `deprecatedInVersion`** corrected from `"3.4.1"` to `"3.5.0"`. There is no 3.4.1 release, and this was the only such value in the vocabularies. Unrelated to #1410, folded in here rather than carrying its own PR.

## Verification

Checked with SPARQL over the authored sources:
- zero remaining references to the four removed quantity kinds, as subject or object
- all 18 `dcterms:isReplacedBy` in the logarithm space resolve to non-deprecated targets
- both generics hold their full unit sets and migrated codes; all three deprecations carry the complete `deprecated` / `deprecatedInVersion` / `isReplacedBy` pattern
- `unit:DECADE` now points only at `CommonLogarithmicFrequencyInterval`


steveraysteveray:
**Update — the "Deliberately NOT merged" section above is superseded.** The PR description at the top only documents the original commit (`4f53f4f8`), which consolidated the `InformationContent`/`InformationFlowRate` families and left the `Ratio`/`FrequencyInterval`/`RatioPerLength`/`TimeRelated` families alone. That part still stands as written.

But an audit of every release since 2.1.45 (Nov 2024) showed `Log10Ratio`, `LogERatio`, `Log10FrequencyInterval` and unqualified `LogarithmicFrequencyInterval` have been stable in every release, coexisting with verbose duplicates (`LogarithmRatioToBase10`/`…BaseE`, `LogarithmicFrequencyIntervalToBase10`) that were also live in 3.4.0 — that duplication, not a missing convention, was the actual inconsistency. So five later commits go further:

- `f55bbfc2` — the short, already-released names (`Log10Ratio`, `LogERatio`, `Log10FrequencyInterval`, `LogarithmicFrequencyInterval`) are kept live instead of being deprecated in favour of new `Common`/`Natural`/`Binary`-prefixed names; the verbose duplicates are deprecated into them instead. `TimeRelatedLogarithmicRatio` → new `TimeRelatedLogERatio`, since its unit (`NP-PER-SEC`) is base-e and the convention now requires that to be explicit.
- `a6fdba12` — `CommonLogarithmicRatioPerLength` (itself unreleased, minted independently in the unrelated, already-merged #1501) is replaced by a new `Log10RatioPerLength` rather than reviving `Linear`/`LineicLogarithmicRatio`, since its own description already asserts base 10 and leaving that implicit would be inconsistent with its siblings.
- `9b51e5fb` — added `Log2FrequencyInterval` as an `exactMatch` synonym of `LogarithmicFrequencyInterval` (base-2 by definition — unit `OCT`, IEC `UAD083` — but not stated in the name), rather than deprecating the released unqualified name. Mirrors how `unit:LB`/`unit:LB_M` coexist as an ambiguous/unambiguous pair.

Net effect: every logarithm-base-carrying quantity kind in the vocabulary now states its base explicitly, except the historically-released `LogarithmicFrequencyInterval`, which gets the disambiguated `Log2FrequencyInterval` alongside it instead of being deprecated. See #1410 for the full closeout against the original issue.

fkleedorfer (approved):

PR #1472, opened 2026-06-01
Update Quantity Kinds to ECLASS Release 16.0

ektrah:
* ECLASS Release 16.0 contains 279 `<unitsml:Quantity />` definitions.
* In QUDT 3.5.0, there are 150 quantity kinds that have an `qudt:eclassCode` assigned.
    * 20 quantity kinds with an `qudt:eclassCode`  are deprecated.
    * 130 quantity kinds with an `qudt:eclassCode` are not deprecated.
    * In total, there are 155 `qudt:eclassCode` statements.
* This PR updates the 155 statements so that they match the respective definition in ECLASS.
* There are some differences between ECLASS and QUDT w.r.t. dimensions:
    * A `<unitsml:Dimension />` definition can include `PlaneAngle` in addition to `AmountOfSubstance`, `ElectricCurrent`, `Length`, `LuminousIntensity`, `Mass`, `ThermodynamicTemperature`, and `Time`.
    * It models the `qudt:dimensionlessExponent` in a different way.
    * It doesn't have `qudt:qkdvNumerator` or `qudt:qkdvDenominator`.

PR #1440, opened 2026-04-24
Replace sh:disjoint with sh:not + sh:class

ektrah:
Unlike `owl:disjointWith`, [`sh:disjoint`](https://www.w3.org/TR/shacl/#DisjointConstraintComponent) expects a predicate, not a class. As a result, `qudt:AuralCue` and other classes are currently inferred to be instances of `rdf:Property` (from the `rdfs:range` of `sh:disjoint`), and the disjointness is probably never checked.

This PR replaces `sh:disjoint` with `sh:not` + `sh:class`:

---

* (Add) An instance of `qudt:AuralCue` must not be an instance of `qudt:IconicCue`, `qudt:KinestheticCue`, or `qudt:VisualCue`
* (Fix) An instance of `qudt:IconicCue` must not be an instance of `qudt:AuralCue`, `qudt:KinestheticCue`, or `qudt:VisualCue`
* (Add) An instance of `qudt:KinestheticCue` must not be an instance of `qudt:AuralCue`, `qudt:IconicCue`, or `qudt:VisualCue`
* (Add) An instance of `qudt:VisualCue` must not be an instance of `qudt:AuralCue`, `qudt:IconicCue`, or `qudt:KinestheticCue`

---

* (Fix) An instance of `qudt:DoublePrecisionType` must not be an instance of `qudt:SinglePrecisionType`
* (Add) An instance of `qudt:SinglePrecisionType` must not be an instance of `qudt:DoublePrecisionType`

---

* (Fix) An instance of `qudt:SignedType` must not be an instance of `qudt:UnsignedType`
* (Fix) An instance of `qudt:UnsignedType` must not be an instance of `qudt:SignedType`

---

* (Fix) An instance of `qudt:EngineeringValueTupleMember` must not be an instance of `qudt:RawValueTupleMember` or `qudt:RawValueTupleMemberType`
* (Add) An instance of `qudt:RawValueTupleMember` must not be an instance of `qudt:EngineeringValueTupleMember` or `qudt:RawValueTupleMemberType`
* (Add) An instance of `qudt:RawValueTupleMemberType` must not be an instance of `qudt:EngineeringValueTupleMember` or `qudt:RawValueTupleMember`

Not sure why `qudt:RawValueTupleMember` and `qudt:RawValueTupleMemberType` are on the same level, but left this in.

---

* (Remove) An instance of `qudt:PositiveBigIntegerType` must not be an instance of `qudt:SignedIntegerType`
[...]

PR #1516, opened 2026-07-29
Stable blank node ordering (Requires Java 25 for build)

fkleedorfer:
Bump maven plugin versions (Requires Java 25+)

- spotless: includes new turtle-formatter for stable blank node
  ordering, requires Java 25
- groovy-plugins: updated for compatibility with Java 25
- reformat TTL sources
- update java versions in github workflows


steveraysteveray:
**Correction below — see [follow-up](https://github.com/qudt/qudt-public-repo/pull/1516#issuecomment-5119501843) for revised CI analysis and conflict list.**

---

I'm willing to work with you on some of this, but the following points need to be addressed.

**1. Migration sequencing for the other open PRs**

Five other PRs are open (#1472, #1440, #1296, #1290, #1220), all branched before this
bump, so they still carry gmavenplus 4.1.1, groovy-ant 4.0.23 and spotless 2.46.1.
Once this merges, anyone reviewing or building those branches has to keep a pre-25 JDK
installed alongside 25 and switch per branch. That's workable, but it should be a
conscious decision rather than a surprise. Can we agree on the order we land these, and
whether the older PRs get rebased promptly after this one?

**2. Three of those PRs will conflict with the reformat**

This PR reformats 12 TTL files. Overlap with the open PRs:

- #1472 (eclass-update) — 1 file, conflicts
- #1440 (fix-disjoint) — 1 file, conflicts
- #1290 (hf-fix_absolute_activity_etc) — 2 files, conflict
- #1296, #1220 — no overlap, clean

For the three affected, hand-resolving blank-node reordering hunks would be error-prone.
I'd suggest we settle the recipe up front: merge main in, keep the branch's semantic
edits, then re-run `mvn spotless:apply` to re-normalise. Worth putting in the PR
description so whoever rebases doesn't invent their own approach.

**3. BUILDING.md is now wrong**

BUILDING.md line 65 still says "Java: Version 11 or higher", and line 88 has
`java -version  # Should show 11 or higher`. This PR updates the workflows but not the
contributor docs, so the first person to follow BUILDING.md after the merge hits the
spotless JVM error with no explanation. Three things to fix:

- Bump the stated Java prerequisite to 25+.
- Add a note to verify with `mvn -v`, not `java -version`. I lost time on exactly this:
  I had JDK 25 installed and `java -version` was beside the point, because Maven was
  still launching 21 from a JAVA_HOME set elsewhere. Anyone with a .mavenrc, a Homebrew
  maven wrapper, jenv or SDKMAN can hit the same split, and the error message ("You are
  running Spotless on JVM 21") doesn't point at the cause.
- Re-check the stated Maven minimum. BUILDING.md says 3.6.0 or higher — does spotless
  3.9.0 still support that? I haven't verified it and don't want to guess.

Happy to take the BUILDING.md changes myself as a follow-up if you'd rather keep this
PR focused.

steveraysteveray:
Correction to my earlier comment. I test-merged the open PRs and checked how the
workflows resolve refs — the CI picture is better than I implied, but the conflict list
is different from what I gave. Revised, in priority order:

**1. BUILDING.md needs updating before this merges**

BUILDING.md line 65 still says "Java: Version 11 or higher", and line 88 has
`java -version  # Should show 11 or higher`. This PR updates the workflows but not the
contributor docs.

I've put this first because it's the only one of the three that's permanent and affects
everyone. The conflicts below resolve themselves as PRs land; the JDK floor is a new
prerequisite for every contributor from here on, and it can't be fixed by a commit — it
lives in each person's JAVA_HOME setup.

It also fails at a distance from its cause. I had JDK 25 correctly installed and still
got "You are running Spotless on JVM 21", because Maven was resolving JAVA_HOME from a
`.mavenrc` I'd forgotten about. Anyone with a Homebrew maven wrapper, jenv, SDKMAN or an
IDE-bundled Maven can hit the same split, and the error text doesn't point at the cause.

Three changes:

- Bump the stated Java prerequisite to 25+.
- Add: verify with `mvn -v`, not `java -version`. Having 25 installed is not sufficient.
- Re-check the stated Maven minimum (currently "3.6.0 or higher") against what spotless
  3.9.0 actually supports. I haven't verified it and don't want to guess.

Happy to take this as a follow-up PR if you'd rather keep this one focused.

**2. Four of the five open PRs will conflict — worth agreeing one recipe**

I test-merged each PR head against this branch with `git merge-tree`:

- #1472 — `CHANGELOG.md` + `VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl`
- #1440 — `CHANGELOG.md` only (its TTL auto-merged cleanly)
- #1290 — `VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl` + `VOCAB_QUDT-UNITS-ALL.ttl`
- #1220 — `pom.xml` (no TTL overlap at all; it collides on the plugin-version block)
- #1296 — clean

Hand-resolving blank-node reordering hunks would be error-prone, so I'd suggest we agree
the approach up front and put it in the PR description: merge main in, keep the branch's
semantic edits, then re-run `mvn spotless:apply` to re-normalise. Note #1220 is a
different kind of conflict and needs a real decision about the pom, not a reformat.

Also worth flagging: merging this won't re-trigger CI on the other PRs, since
`pull_request` runs fire on push, not when the base branch moves. Their green checks will
be stale — worth a manual re-run on each before merging.

**3. Two workflows will run old plugins under JDK 25**

I was wrong that the PR CI is at risk — `maven.yml` builds `refs/pull/N/merge`, so pom.xml
and the workflow file always travel together and can't disagree about the JDK. That path
is fine.

But `compare-dist.yml` and `manual-validation-build.yml` check out a named branch
(`ref: inputs.head_branch` and `ref: inputs.branch || github.ref`), not a merge ref. Once
this lands, dispatching either from main uses main's workflow file (JDK 25) against the
target branch's own source, which still has spotless 2.46.1 and groovy-ant 4.0.23. That
combination is exactly the mismatch this PR is fixing. It only bites on un-rebased
branches, so it's transient, but it'll be confusing while it lasts.

fkleedorfer:
Given that the real changes in this PR are limited to pom and workflow yml files - all the touched ttl files result from formatting - I suggest first merging any open PRs and then rebasing this one, (minus its changes to ttls), reformatting and merging it

steveraysteveray:
I think we will need to handle each of the open PRs on a case-by-case basis:

1220: That's your preliminary approach for profiles. Suggest we close that for now.
1290: The premise of this PR has changed considerably, especially in 3.5.0, so we should reach out to @henrikef and suggest we close this PR and revisit if necessary. By mentioning her here, I think GitHub will be notifying her.
1296: This is yours. Do you still want to pursue it?
1440: This is the one that @ralphtq has been addressing in his datatypes updates, even though the original issue arose out of sh:disjoint. It's not ready to merge yet. I believe the work by @ralphtq is going far beyond the original issue raised by @ektrah and subsumes that issue, so we could probably close this one and @ralphtq could open a new PR when he is ready, in which case it will be using the new blank ordering anyway
1472: This one looks good to merge, but will first need the merge conflicts resolved. I would ask @ektrah to do that, and then we can approve and merge.

So, bottom line, as you suggest, we merge 1472 but close all the others, with the agreement of the involved parties. We will need to notify them also about upgrading their local Maven environments.

We also still need to address items 1 and 3 in my earlier comment.

steveraysteveray:
A heads-up on sequencing, and a request for a review elsewhere.

I've picked up #1290 (open since September 2025), which corrects the dimension vector of `quantitykind:AbsoluteActivity` from `qkdv:A0E0L-3I0M0H0T0D0` to dimensionless.

I merged `main` into Henrike's branch, resolved ten months of drift, and fixed several defects (a doubled prefix, two Turtle syntax errors, a 10× error in `conversionMultiplierSN`, and two `skos:broader` links that should be `qudt:organizedUnder` under the dimensionless-decoupling convention). It is now MERGEABLE and the build is clean.

**Why this concerns #1516:** #1290 touches both `VOCAB_QUDT-QUANTITY-KINDS-ALL.ttl` and `VOCAB_QUDT-UNITS-ALL.ttl` — the same files this PR re-serialises. If #1516 lands first, #1290 acquires blank-node reordering conflicts on top of the semantic ones we just cleared. So I'd like to land #1290 first, then this.

#1290 is currently blocked on codeowner review. Would one of you take a look so we can merge it ahead of this one? It's a small diff and I've summarised every change in a comment on that PR.

steveraysteveray (approved):

PR #1523, opened 2026-08-12
Add days per year as a unit

Maximvdw:
## Summary

Adds `unit:DAY-PER-YR` — "Day per Year" (NL: "Dag per Jaar") — to
`vocab/unit/VOCAB_QUDT-UNITS-ALL.ttl`.

Day per year expresses the ratio of the number of days to the number of years
during which an event or process occurs within one year. It is modeled on the
existing `unit:HR-PER-YR` (Hour per Year):

**Motivation/use case:** For the [Industrial Emission Portal Regulation](https://environment.ec.europa.eu/topics/industrial-emissions-and-safety/industrial-emissions-portal-regulation-iepr_en) we require a unit to express how many days per year emissions occur from a particular point. Other use cases include for example 'amount of rain in a year: days/year'.

- `a qudt:Unit`
- dimensionless: `qudt:hasDimensionVector qkdv:A0E0L0I0M0H0T0D1`
- `qudt:unitForQuantityKind quantitykind:TimeRatio`
- `qudt:conversionMultiplier 0.002737850787132101300479123887748118` (= 86400 s / 31557600 s, i.e. DAY / YR)
- `qudt:symbol "d/a"`, `qudt:ucumCode "d.a-1"`

The singular qname `DAY-PER-YR` follows the naming conventions
(`DAY-PER-NUM`, `HR-PER-YR`) and avoids a spurious contextual `DAYS` unit that
the build would otherwise auto-create for the plural form.

## Changelog

- Added `unit:DAY-PER-YR`, a dimensionless unit expressing days per year (`TimeRatio`), symbol `d/a` ([Maxim Van de Wynckel](https://github.com/Maximvdw)).

## Checklist

- [x] Built locally with `mvn install` (build success, 0 violations)
- [x] Added entry to `CHANGELOG.md`


steveraysteveray (approved):

jhodgesatmb:
Would days per year (Gregorian calendar) be the inverse of the conversion multiplier here? At least, when I think of days per year I think 365, not .0273... More importantly, it isn't a constant, and it doesn't represent the lunar calendar. So this ratio is specific but not specified as such. Isn't that a problem?

steveraysteveray:
I think the desired use here would be interpreted as "days of interest per year" or something like that. Not the definition of a year in days. Things like "days experiencing power outage each year", etc.

PR #1524, opened 2026-08-12
Add ToxicEquivalent and TEQ units

Maximvdw:
## Summary

Adds `quantitykind:ToxicEquivalent` and the units `unit:TEQ` and
`unit:TEQ-PER-YR`.

**Motivation/use case**: For the [Industrial Emissions Portal Regulation](https://environment.ec.europa.eu/topics/industrial-emissions-and-safety/industrial-emissions-portal-regulation-iepr_en) we require reporting of the toxic equivalent per year for particular emissions.

Toxic equivalent (TEQ) expresses the toxicity-weighted mass of dioxin-like
compounds (PCDDs, PCDFs, PCBs) relative to the most toxic congener
2,3,7,8-TCDD: `TEQ = Σ[Cᵢ × TEFᵢ]`. It is used for dioxin emission
reporting in air, e.g. `0.05 ng TEQ/Nm³` and `20 mg TEQ/jaar`.

> Note: TEQ/Nm³ is currently not part of this PR but is given as an example

- `quantitykind:ToxicEquivalent` — `specializationOf quantitykind:MassEquivalent`, dimension M (mirrors `quantitykind:CO2Equivalent`)
- `unit:TEQ` — `ContextualUnit`, 1 TEQ = 1 gram of toxic equivalent (`hasFactorUnit unit:GM`, `skos:broader unit:GM`), quantity kinds `Mass` and `ToxicEquivalent` (mirrors `unit:GM_Carbon`)
- `unit:TEQ-PER-YR` — annual toxic equivalent mass flow, dimension M/T, `quantitykind:MassFlowRate`, `symbol "g{TEQ}/a"`, `ucumCode "g.a-1{TEQ}"` (mirrors `unit:KiloGM-PER-YR`)

## Changelog

- Added `quantitykind:ToxicEquivalent` (`specializationOf quantitykind:MassEquivalent`) with the new `unit:TEQ` (one TEQ = one gram of toxic equivalent, `ContextualUnit`) and `unit:TEQ-PER-YR` for annual toxic equivalent emissions ([Maxim Van de Wynckel](https://github.com/Maximvdw)).

## Checklist

- [x] Built locally with `mvn install` (build success, 0 violations)
- [x] Added entry to `CHANGELOG.md`


fkleedorfer:
Please don't use a single `qudt:hasFactorUnit` (that form is intended only to be used with qudt:factorUnitScalar. see e.g., unit:PINT_UK, which is half a quart, apparently). Replace the factor unit with `unit:TEQ qudt:scalingOf unit:KiloGM`.

EDIT: on second thought - let's keep it like this. GM_DRY works the same way.

fkleedorfer:
I defer to others for their assessment of the new quantitykind (@steveraysteveray?)

steveraysteveray (commented):
I'm no expert here, but in reading your reference to Wikipedia, it seems like the relevant quantity kind would be a dimensionless concept, ToxicEquivalencyFactor, rather than a mass-based concept. Similarly, the TEQ unit would be dimensionless. Again, I defer to your familiarity with how these are used in the community, but is this worth considering?

Maximvdw:
Toxic Equivalence Factor (TEF) is used to calculate the Toxic Equivalent [[1](https://www.epa.gov/toxics-release-inventory-tri-program/dioxin-and-dioxin-compounds-toxic-equivalency-information)][[2](https://www.bundesumweltministerium.de/en/topics/health/food-safety/consumer-protection/dioxins-and-pcbs/toxicity-equivalents-for-dioxins)]

It is calculated using the mass/concentration (multiplication) so while TEF is dimensionless, the TEQ would be mass-based in my opinion 
TEQ = Σ[Ci × TEFi]
> The toxic equivalency of a mixture is defined by the sum of the concentrations of individual compounds (Ci) multiplied by their relative toxicity (TEF):[[6]](https://en.wikipedia.org/wiki/Toxic_equivalency_factor#cite_note-Safe-6)

In addition; reporting often combines weight + TEQ so it would be:

"mg TEQ/year" (Dutch site: https://iplo.nl/regelgeving/regels-voor-activiteiten/emissie-zzs-lucht/)
image

Maximvdw:
Tagging @gezever as he might also have more knowledge about this particular topic/unit

steveraysteveray:
OK, I'm now seeing your point of view better. You are modeling the quantity kind much the same way we modeled CO2Equivalent. However, as in the case of CO2Equivalent, we don't have dedicated units - they are just plain KiloGM or TONNE, units of mass, with the quantitykind setting the context. When people say "the emissions were 250 Kilograms of CO2 equivalent", the unit is Kilgrams, the quantity kind (i.e. the context) is CO2Equivalent. People conflate the unit with the quantity kind. 

So bottom line, I'm concluding now that all you need is your quantity kind, and model it as a sibling of CO2Equivalent, which is exactly what you did. You can then narrow which mass units make sense in your field, just as we did with CO2Equivalent with the qudt:unitForQuantityKind relation.
image

Does that work for you?

steveraysteveray:
One last thought is that the only new unit needed would be unit:GM-PER-YR, which would also be available for others in different contexts, with an associated quantity kind of MassFlowRate (already defined).

Maximvdw:
Hi, thank you for the feedback - I will have to investigate this a bit since I have not looked at CO2Equivalent yet.

steveraysteveray:
@Maximvdw, you might also find [this documentation](https://github.com/qudt/qudt-public-repo/wiki/Commensurability-Composition-Semantics-and-Context#4-worked-example-equivalents-chemistry) in our wiki helpful. The topic of equivalent units has also been discussed in #1484, #1343, #882. In reviewing some of these older PRs, I am reminded that we have tried to be flexible when a community of practice commonly uses an Equivalent measure as a unit. That is what we ultimately did for the user in #882 who asked for unit:EQ. So to be fair, if you really want unit:TEQ, we should handle it the same. The trade-off, aside from conflating the distinction between a unit and a quantity kind, is that you would need to define all compound units using your TEQ unit, such as unit:TEQ-PER-\<whatever\>, as contrasted with simply using the already defined underlying mass-based units.

steveraysteveray:
@Maximvdw, checking in to see if you have come to a conclusion on this PR.

PR #1525, opened 2026-08-21
made changes to the unit vocab as per G. Rossi to harmonize with IEC CDD

jhodgesatmb:
Received a request to make some changes to the Unit vocabulary to harmonize with the IEC CDD. Changes included removed of some iec61360Code values and some UNECE Common Code values.


jhodgesatmb:
I forgot to run the established workflows before creating the PR.

